PR #6, opened 2025-09-17
make kafka client ids unique by hostname

haileyok:
Generally, client IDs should be unique. There's some different naming patterns for doing this that I've seen:

- An "app name" and use a "client id" as something unique
- A "client id" and use a "suffix" as something unique

I have no preference, but here I just kind of with with [what we do](https://github.com/bluesky-social/go-util/blob/main/pkg/bus/consumer/consumer.go#L113-L118) while trying to remain flexible with it... There's also a question of "do we really want to use hostnames for this". We do, since each pod in a k8s cluster will have a unique hostname. Some people though use UUIDs for this which would also be acceptable.

Notably though, if these are not unique (and - maybe - if the server isn't configured correctly? I'm not really a kafkaer) multiple consumers may receive the same events.

ps:  is there a readme or something that explains all the various env vars? or are we just putting them in the compose?

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -77,6 +77,8 @@ services:
       - OSPREY_KAFKA_BOOTSTRAP_SERVERS=["kafka:29092"]
       - OSPREY_KAFKA_INPUT_STREAM_TOPIC=osprey.actions_input
       - OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID=localhost
+      # Leave missing to automatically use hostname for client id suffix
+      # - OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID_SUFFIX=
       - OSPREY_KAFKA_OUTPUT_SINK=True
       - OSPREY_KAFKA_OUTPUT_TOPIC=osprey.execution_results
       - OSPREY_KAFKA_OUTPUT_CLIENT_ID=localhost
--- a/osprey_worker/src/osprey/worker/sinks/__init__.py
+++ b/osprey_worker/src/osprey/worker/sinks/__init__.py
@@ -8,6 +8,8 @@ Therefore we patch at the top of this file -- it's generally fine to double patc
 but not ideal as clients should be responsible for patching
 """
 
+import platform
+
 from osprey.worker.lib.patcher import patch_all
 
 patch_all()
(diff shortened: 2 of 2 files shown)


vinaysrao1:
@julietshen - should we tag the release version before this PR is merged?

osprey_worker/src/osprey/worker/sinks/__init__.py
```diff
@@ -143,10 +145,18 @@ def get_rules_sink_input_stream(
     elif input_stream_source == InputStreamSource.KAFKA:
         config = CONFIG.instance()
         client_id = config.get_str('OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID', 'localhost')
+        client_id_suffix = config.get_optional_str('OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID_SUFFIX')
         input_topic: str = config.get_str('OSPREY_KAFKA_INPUT_STREAM_TOPIC', 'osprey.actions_input')
         input_bootstrap_servers: list[str] = config.get_str_list('OSPREY_KAFKA_BOOTSTRAP_SERVERS', ['localhost'])
         group_id = config.get_optional_str('OSPREY_KAFKA_GROUP_ID')
 
+        # If a suffix isn't provided, make sure that client IDs are unique by hostname
+        if client_id_suffix is None:
```
EXBreder: Maybe just have the client_id default to the hostname, instead of having the hostname replace the suffix?

And if the suffix is unset, the client_id just remains the client_id. This way the client_id can be explicitly set instead of modified under the hood.

haileyok: okay think i understood that right, though do verify

osprey_worker/src/osprey/worker/sinks/__init__.py
```diff
@@ -142,11 +144,16 @@ def get_rules_sink_input_stream(
...
     elif input_stream_source == InputStreamSource.KAFKA:
         config = CONFIG.instance()
-        client_id = config.get_str('OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID', 'localhost')
+        client_id = config.get_str('OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID', platform.node())
+        client_id_suffix = config.get_optional_str('OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID_SUFFIX')
         input_topic: str = config.get_str('OSPREY_KAFKA_INPUT_STREAM_TOPIC', 'osprey.actions_input')
         input_bootstrap_servers: list[str] = config.get_str_list('OSPREY_KAFKA_BOOTSTRAP_SERVERS', ['localhost'])
         group_id = config.get_optional_str('OSPREY_KAFKA_GROUP_ID')
 
+        # If a suffix isn't provided, make sure that client IDs are unique by hostname
```
EXBreder: yup this is good!

docker-compose.yaml
```diff
@@ -77,6 +77,8 @@ services:
       - OSPREY_KAFKA_BOOTSTRAP_SERVERS=["kafka:29092"]
       - OSPREY_KAFKA_INPUT_STREAM_TOPIC=osprey.actions_input
       - OSPREY_KAFKA_INPUT_STREAM_CLIENT_ID=localhost
+      # Leave missing to automatically use hostname for client id suffix
```
EXBreder: "suffix" is not necessary anymore

haileyok: yep, going to remove suffix entirely from here in that case and we can put this somewhere that the rest of vars get documented at later

haileyok: (and a comment to client_id saying if you leave it out it'll default to hostname)

PR #9, opened 2025-09-17
feat: registerable input stream

haileyok:


entrypoint.sh
```diff
@@ -32,7 +32,7 @@ cli-osprey-ui-api() {
 }
 
 cli-osprey-worker() {
-    exec uv run python3.11 osprey_worker/src/osprey/worker/cli/sinks.py run-rules-sink --input kafka --rules-path ./example_rules
+    exec uv run python3.11 osprey_worker/src/osprey/worker/cli/sinks.py run-rules-sink --input custom --rules-path ./example_rules
```
haileyok, line 35: this could just be an env var :(

PR #10, opened 2025-09-17
Input Stream Plugin

EXBreder:
Adds an input stream plugin mechanism for defining separate plugins for input strteams that aren't just the defacto kafka/pubsub/stdin.

This also enables customization of existing input stream beyond the standard env var settings.

### DIFF
--- a/osprey_worker/src/osprey/worker/cli/sinks.py
+++ b/osprey_worker/src/osprey/worker/cli/sinks.py
@@ -115,20 +115,6 @@ def tail_kafka_input_sink() -> None:
 
 @cli.command()
 @click.option(
-    '--input',
-    'input_stream_source',
-    default=InputStreamSource.PUBSUB,
-    type=EnumChoice(InputStreamSource),
-    help='Where to ingest data from.',
-)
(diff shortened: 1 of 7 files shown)


osprey_worker/src/osprey/worker/cli/sinks.py
```diff
@@ -114,20 +114,6 @@ def tail_kafka_input_sink() -> None:
 
 
 @cli.command()
-@click.option(
```
EXBreder, line 117: this option was not used anymore

osprey_worker/src/osprey/worker/sinks/input_stream_chooser.py
```diff
@@ -0,0 +1,122 @@
+import random
```
EXBreder, line 2: Moved files, avoids some nasty circular import logic

haileyok (approved):
nice, works on my machine (merged into my branch, works great locally)

BinaryFiddler (approved):

haileyok (approved):

PR #15, opened 2025-09-19
Add an output sink writing results to minio, pluggify storage option

BinaryFiddler:
- Add an output sink that stores full execution results
- Add a rule in the example to populate the ActionName
- Plugify storage option for execution result
- Give super user access to view action and feature data

Screenshot 2025-09-24 at 4 40 59 PM

### DIFF
--- a/example_rules/main.sml
+++ b/example_rules/main.sml
@@ -1,2 +1,3 @@
+Import(rules=['models/base.sml'])
+
 Require(rule='rules/post_contains_hello.sml')
-ActionName=GetActionName()
--- a/example_rules/main.sml
+++ b/example_rules/main.sml
@@ -1,1 +1,2 @@
 Require(rule='rules/post_contains_hello.sml')
+ActionName=GetActionName()
--- a/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
+++ b/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
@@ -1,7 +1,7 @@
 from typing import List, Sequence
 
 from kafka import KafkaProducer
-from osprey.worker.adaptor.plugin_manager import hookimpl_osprey
+from osprey.worker.adaptor.plugin_manager import hookimpl_osprey, bootstrap_execution_result_store
 from osprey.worker.lib.config import Config
 from osprey.worker.sinks.sink.kafka_output_sink import KafkaOutputSink
 from osprey.worker.sinks.sink.output_sink import BaseOutputSink, StdoutOutputSink
@@ -23,5 +23,7 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
                 kafka_producer=KafkaProducer(bootstrap_servers=bootstrap_servers, client_id=client_id),
             )
         )
-    sinks.append(StoredExecutionResultOutputSink())
+    execution_result_store = bootstrap_execution_result_store(config=config)
+    if execution_result_store is not None:
+        sinks.append(StoredExecutionResultOutputSink())
     return sinks
--- a/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
+++ b/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
@@ -7,6 +7,7 @@ from osprey.worker.sinks.sink.kafka_output_sink import KafkaOutputSink
 from osprey.worker.sinks.sink.output_sink import BaseOutputSink, StdoutOutputSink
 from osprey.worker.sinks.sink.stored_execution_result_output_sink import StoredExecutionResultOutputSink
 
+
 @hookimpl_osprey
 def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
     sinks: List[BaseOutputSink] = []
--- a/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
+++ b/osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
@@ -5,7 +5,7 @@ from osprey.worker.adaptor.plugin_manager import hookimpl_osprey
 from osprey.worker.lib.config import Config
 from osprey.worker.sinks.sink.kafka_output_sink import KafkaOutputSink
 from osprey.worker.sinks.sink.output_sink import BaseOutputSink, StdoutOutputSink
-
+from osprey.worker.sinks.sink.stored_execution_result_output_sink import StoredExecutionResultOutputSink
 
 @hookimpl_osprey
 def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
@@ -22,4 +22,5 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
                 kafka_producer=KafkaProducer(bootstrap_servers=bootstrap_servers, client_id=client_id),
             )
         )
+    sinks.append(StoredExecutionResultOutputSink())
     return sinks
--- a/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
+++ b/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
@@ -48,7 +48,7 @@ def register_input_stream(config: Config) -> BaseInputStream[BaseAckingContext[A
     raise NotImplementedError('register_input_stream must be implemented by the plugin')
 
 
-@hookspec
+@hookspec(firstresult=True)
 def register_execution_result_store(config: Config) -> ExecutionResultStore:
     """Register an execution result storage backend instance."""
     raise NotImplementedError('register_execution_result_store must be implemented by the plugin')
--- a/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
+++ b/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
@@ -13,6 +13,7 @@ from osprey.worker.sinks.utils.acking_contexts import BaseAckingContext
 
 if TYPE_CHECKING:
     from osprey.worker.lib.config import Config
+    from osprey.worker.lib.storage.stored_execution_result import ExecutionResultStore
     from osprey.worker.sinks.sink.output_sink import BaseOutputSink
 
 hookspec: pluggy.HookspecMarker = pluggy.HookspecMarker(OSPREY_ADAPTOR)
@@ -45,3 +46,8 @@ def register_action_proto_deserializer() -> ActionProtoDeserializer | None:
 @hookspec(firstresult=True)
 def register_input_stream(config: Config) -> BaseInputStream[BaseAckingContext[Action]]:
     raise NotImplementedError('register_input_stream must be implemented by the plugin')
+
+@hookspec
+def register_execution_result_store(config: Config) -> ExecutionResultStore:
+    """Register an execution result storage backend instance."""
+    raise NotImplementedError('register_execution_result_store must be implemented by the plugin')
--- a/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
+++ b/osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
@@ -47,6 +47,7 @@ def register_action_proto_deserializer() -> ActionProtoDeserializer | None:
 def register_input_stream(config: Config) -> BaseInputStream[BaseAckingContext[Action]]:
     raise NotImplementedError('register_input_stream must be implemented by the plugin')
 
+
 @hookspec
 def register_execution_result_store(config: Config) -> ExecutionResultStore:
     """Register an execution result storage backend instance."""
--- a/osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
+++ b/osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
@@ -101,4 +101,4 @@ def bootstrap_execution_result_store(config: Config):
         [store] = plugin_manager.hook.register_execution_result_store(config=config)
         return store
     except Exception as e:
-        raise RuntimeError(f'No storage backend plugin found: {e}')
+        raise RuntimeError(f'Storage backend plugin could not be loaded: {e}')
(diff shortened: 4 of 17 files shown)


example_rules/main.sml
```diff
@@ -1 +1,2 @@
 Require(rule='rules/post_contains_hello.sml')
+ActionName=GetActionName()
```
haileyok: smol nit, i'd probably put `ActionName` in `models/base.sml` to keep with that pattern, then import `base.sml` in here. Good as an example on how imports work.

```suggestion
# Import base variables and entities
Import(rules=['models/base.sml'])

Require(rule='rules/post_contains_hello.sml')
```

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -67,38 +90,36 @@ def persist_from_execution_result(cls, execution_result: ExecutionResult) -> Non
...
-                StoredExecutionResult.parse_from_query_result(result, data_censor_abilities)
-                for result in combined_results
-            ],
+            [StoredExecutionResult.parse_from_query_result(result, data_censor_abilities) for result in results],
             key=lambda r: pytz.utc.localize(r.timestamp) if r.timestamp.tzinfo is None else r.timestamp,
             reverse=True,
         )
 
+    @classmethod
+    def _get_storage_backend(cls) -> ExecutionResultStore:
```
EXBreder: Perhaps invert this - have another component set the storage backend instead of having this grab the bootstrap stuff directly. With https://github.com/roostorg/osprey/pull/18, its only one place that this would need to be called

EXBreder: By inverting this, the stored execution result doesn't need to think about how configuration is generated

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -324,17 +335,115 @@ def insert(
...
+                    'id': action_id,
+                    'extracted_features': extracted_features_json,
+                    'error_traces': error_traces_json,
+                    'timestamp': timestamp.isoformat(),
+                    'action_data': action_data_json,
+                }
+
+                json_data = json.dumps(data)
+
+                from io import BytesIO
```
EXBreder: why this import at runtime?

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -324,17 +335,115 @@ def insert(
...
+                    raise
+
+        except Exception as e:
+            logger.error(f'Failed to retrieve execution result from MinIO for action_id {action_id}: {e}')
+            return None
+
+    def select_many(self, action_ids: List[int]) -> List[Dict[str, Any]]:
+        results = [
+            result
+            for result in gevent.pool.Pool(GCS_CONCURRENCY_LIMIT).imap(self.select_one, action_ids)
```
EXBreder: Should rename this concurrency limit if its not gcs

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
BinaryFiddler: Made a thin wrapper around the `StoredExecutionResult` since itself is a pydantic model.

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
```diff
@@ -22,4 +23,5 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
                 kafka_producer=KafkaProducer(bootstrap_servers=bootstrap_servers, client_id=client_id),
             )
         )
+    sinks.append(StoredExecutionResultOutputSink())
```
jaredmiller13: Do we want this behind a flag like the other outputs?

osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
```diff
@@ -91,3 +91,14 @@ def bootstrap_input_stream(config: Config) -> BaseInputStream[BaseAckingContext[
...
+
+def bootstrap_execution_result_store(config: Config):
+    """Get the execution result storage backend from plugins."""
+    load_all_osprey_plugins()
+
+    try:
+        [store] = plugin_manager.hook.register_execution_result_store(config=config)
+        return store
+    except Exception as e:
+        raise RuntimeError(f'No storage backend plugin found: {e}')
```
jaredmiller13: This error message isn't accurate right? There might be a plugin but it failed to load?

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -23,9 +27,35 @@
 if TYPE_CHECKING:
     from osprey.worker.ui_api.osprey.lib.abilities import DataCensorAbility
 
-SCYLLA_CONCURRENCY_LIMIT = 100
 BIGTABLE_CONCURRENCY_LIMIT = 100
 GCS_CONCURRENCY_LIMIT = 100
+MINIO_CONCURRENCY_LIMIT = 100
+
+
+class ExecutionResultStore(ABC):
```
jaredmiller13, line 36: nit: want to make this a protocol instead of an ABC? ABC is better IMO if you're providing some default implementation, which you don't seem to be doing here.

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -159,70 +177,64 @@ def _censor_data(
...
         row.set_cell('execution_result', b'extracted_features', extracted_features_json.encode(), timestamp=timestamp)
         row.set_cell('execution_result', b'error_traces', error_traces_json.encode(), timestamp=timestamp)
         row.set_cell('execution_result', b'timestamp', timestamp.isoformat().encode(), timestamp=timestamp)
         row.set_cell('execution_result', b'action_data', action_data_json.encode(), timestamp=timestamp)
-        osprey_bigtable.table('stored_execution_result').mutate_rows([row], retry=cls.retry_policy)
+        osprey_bigtable.table('stored_execution_result').mutate_rows([row], retry=self.retry_policy)
 
-    @classmethod
-    def _encode_action_id(cls, action_id_snowflake: int) -> bytes:
+    def _encode_action_id(self, action_id_snowflake: int) -> bytes:
```
jaredmiller13: Why not make these static methods instead?

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -324,17 +332,113 @@ def insert(
...
+                self._minio_client.put_object(
+                    self._bucket_name,
+                    object_name,
+                    data_stream,
+                    length=len(json_data.encode('utf-8')),
+                    content_type='application/json',
+                )
+
+        except Exception as e:
+            logger.error(f'Failed to insert execution result into MinIO for action_id {action_id}: {e}')
```
jaredmiller13, line 433: Do we want to add any sort of handling if the insert fails? Like a retry or anything?

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -324,17 +332,113 @@ def insert(
...
+
+    def _encode_action_id(self, action_id_snowflake: int) -> str:
+        """Constructs a MinIO object key for a given snowflake using the same distribution logic as BigTable."""
+        timestamp_portion = action_id_snowflake >> 22
+        # reverse the last 4 characters of the timestamp to create a
+        # uniformly distributed prefix space.
+        key_prefix = str(timestamp_portion)[:-5:-1]
+        return f'{key_prefix}:{action_id_snowflake}.json'
```
jaredmiller13: if we don't already have a snowflake util, let's add one. This is the kind of stuff that should be done once to reduce the changes of it being messed up.

osprey_worker/src/osprey/worker/ui_api/osprey/views/events.py
```diff
@@ -19,6 +21,8 @@
 from pydantic.main import BaseModel
 
 from ..lib.auth import get_current_user
+
+
```
jaredmiller13: intentional?

BinaryFiddler: nope

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
BinaryFiddler: I think this is fine for the purpose of a v1, the minio storage option is meant to be a minimal working example anyway. 

We could wrap it in tenacity to get retry/backoff in a separate PR if we decided to.

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
BinaryFiddler: They can be.

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
BinaryFiddler: What makes a protocol more appealing in this case? Protocol also allows providing default implementation, the only difference is we don't have to explicitly sub classing, which I'd argue explicit sub classing from ABC is more desirable here.

osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
BinaryFiddler: I'll make it more general

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
BinaryFiddler: I think we do want to always include one, no? Without it part of UI won't work.

osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
EXBreder: you check if the plugin exists just by checking `plugin_manager.hook.register_execution_result_store` exists, according to pluggy

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
BinaryFiddler: Pulling this one into the `ABC` since it's shared across all 3.

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
EXBreder: I'd like to see the output sink is appended if and only if the plugin is defined

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
EXBreder: Agree that ABCs are better - I like the explicit nature here. 

Protocols in particular are really good if you are trying to integrate other libraries, but this isn't the case here.

jaredmiller13: Protocols allow for default implementation but it goes against the spirit of them. You can subclass a protocol though for what it's worth, the same as an ABC. 

> Protocols in particular are really good if you are trying to integrate other libraries, but this isn't the case here.

Isn't this meant to define how other plugins can integrate their own ExecutionResultStore?

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
jaredmiller13: It's a nit for sure, but I saw you went through the trouble of moving them from a class method to a instance method and wasn't sure why.

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
BinaryFiddler: That's reasonable, I will make that happen

osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
```diff
@@ -45,3 +46,9 @@ def register_action_proto_deserializer() -> ActionProtoDeserializer | None:
 @hookspec(firstresult=True)
 def register_input_stream(config: Config) -> BaseInputStream[BaseAckingContext[Action]]:
     raise NotImplementedError('register_input_stream must be implemented by the plugin')
+
+
+@hookspec
```
EXBreder: https://pluggy.readthedocs.io/en/stable/#first-result-only since this can really only be used once. Note in the implementation usage, you will no longer have to extract it from an array either.

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
EXBreder: My meaning was in more about if you don't have control of a library you are trying to integrate with. In this case, _osprey_ is the library.

And yes, while you _can_ subclass protocols, ABCs make it explicit that you _have_ to.

Regardless, I do think this is a bikeshed. The difference between these two for this particular use cases is really limited.

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -23,9 +27,55 @@
...
+        extracted_features_json: str,
+        error_traces_json: str,
+        timestamp: datetime,
+        action_data_json: str,
+    ) -> None:
+        """Insert an execution result."""
+        pass
+
+    @staticmethod
+    def _generate_key_prefix_from_snowflake(action_id_snowflake: int) -> str:
```
EXBreder: This feels super awkward to keep in here - the action_id_snowflake isn't something that the implementor of this class needs to worry about. Instead - maybe use the osprey.worker.lib.snowflake class and put something there?

osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
BinaryFiddler: that's a cool trick

EXBreder (approved):

PR #19, opened 2025-09-23
add labels provider, remove labels pb2

EXBreder:
doing my best to remove more code.

This PR removes all label protobuf code, generated and otherwise, and replaces all instances with the "local" version. The rationale is that the labels provider, which is now provided by plugins.

This one does _not_ provide a label provider, but that can come next. Its recommend that both a local label provider (with stuff stored in memory) and a postgres version for more (slightly) production flows.

Notable changes include:
- in the plugin bootstrap scripts, the labels UDF and labels output sink is only added if the labels provider is actually there.
- EventEffectOutputSink is just LabelOutputSink now, since it no long handles the webhooks (which also makes it a lot simpler). Its signature is altered, and calling things no longer need the engine for the sink.
- LabelStatus is moved to the engine subdirectory, making one fewer "backwards" dependency. Its also swapped to an intenum for some ease of use.

### DIFF
--- a/osprey_worker/src/osprey/engine/language_types/labels.py
+++ b/osprey_worker/src/osprey/engine/language_types/labels.py
@@ -32,6 +32,13 @@ class LabelStatus(IntEnum):
             case LabelStatus.REMOVED | LabelStatus.MANUALLY_REMOVED:
                 return LabelStatus.REMOVED
 
+    def is_automatic(self) -> bool:
+        """Checks if the status is automatic (ADDED or REMOVED)."""
+        return self in [LabelStatus.ADDED, LabelStatus.REMOVED]
+
+    def is_manual(self) -> bool:
+        """Checks if the status is manual (MANUALLY_ADDED or MANUALLY_REMOVED)."""
+        return self in [LabelStatus.MANUALLY_ADDED, LabelStatus.MANUALLY_REMOVED]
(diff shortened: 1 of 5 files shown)


osprey_worker/src/osprey/worker/sinks/sink/bulk_label_sink.py
```diff
@@ -14,14 +13,17 @@
...
-from osprey.worker.lib.storage.labels import get_for_entity
 from osprey.worker.sinks.sink.input_stream import BaseInputStream
-from osprey.worker.sinks.sink.output_sink import EventEffectsOutputSink
+from osprey.worker.sinks.sink.output_sink import LabelOutputSink
 from osprey.worker.sinks.sink.output_sink_utils.constants import MutationEventType
 from osprey.worker.sinks.sink.output_sink_utils.models import OspreyBulkJobAnalyticsEvent
 from osprey.worker.ui_api.osprey.lib.druid import PeriodData, TopNDruidQuery, TopNPoPResponse
 from tenacity import RetryCallState, retry, retry_if_exception_type, stop_after_attempt, wait_exponential
 
+from ...adaptor.plugin_manager import bootstrap_label_provider
```
EXBreder, line 23: remove relative imports

PR #22, opened 2025-10-01
fix: all engine module tests

caidanw:


osprey_worker/src/osprey/engine/ast_validator/tests/test_validation_context/test_validation_transforms_osprey_py_ast_error.txt
```diff
@@ -2,4 +2,4 @@ error: functions with non-keyword arguments are not supported
 --> main.sml:1:0
      |
    1 | Foo(1)
-     | ^
```
EXBreder, line 5: there are a million of these, and I don't see the difference (doesnt look like a newline delimeter change either). What are they?

osprey_worker/src/osprey/engine/executor/udf_execution_helpers.py
```diff
@@ -28,13 +28,22 @@ def accessor_get(self, execution_context: ExecutionContext, key: KeyT, lock: boo
...
-    """Indicates that a UDF requires a helper object to execute, which will be provided via the execution context."""
+    """Indicates that a UDF requires a helper object to execute, which will be provided via the execution context.
+
+    Subclasses may optionally implement `create_provider` to allow automatic helper bootstrapping via plugins.
+    Tests or callers can also inject helpers explicitly using `UDFHelpers.set_udf_helper` without requiring this method.
+    """
 
     @classmethod
-    @abstractmethod
     def create_provider(cls) -> HelperT:
```
EXBreder, line 35: Why remove the abstractmethod here? it was explicitly that way so users are forced to write the implementor.

osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_experiments.py
```diff
@@ -61,173 +57,17 @@ def test_experiment_bucketing(execute: ExecuteFunction) -> None:
         str(0),
         str(1),
         str(1),
-        str(True),
     ]
 
 
-# TODO: related to ExperimentsProvider import issue above, re-enable when that is resolved
```
EXBreder, line 68: you can disable the test using `@pytest.mark.skip(reason="things")`

osprey_worker/src/osprey/engine/ast_validator/tests/test_validation_context/test_validation_transforms_osprey_py_ast_error.txt
caidanw: It's from when we previously formatted the entire code base. These text files are strict in that they need to match exactly, down to the trailing new lines. So these all just contain fixes for them to match.

osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_experiments.py
caidanw: The `ExperimentsProvider` class no longer exists. So these tests have been removed as there doesn't seem to be a way to update the tests without re-adding the class.

If we just skip this test (without commenting out this code), then this file will throw an import error due to the missing class.

osprey_worker/src/osprey/engine/executor/udf_execution_helpers.py
caidanw: fixed

osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_experiments.py
caidanw: Okay, I revisited this today and was able to keep the tests while fixing the problem.

osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_strings.py
```diff
@@ -182,17 +182,17 @@ class Scenario:
         Scenario('🔞t3xt'),
         Scenario('ΠOPQ 100-', ['nop', '100']),
         Scenario('Te.xt'),
-        Scenario('∏opq'),
+        Scenario('∏opq', 'nopq'),
```
BinaryFiddler, line 185: What does the second param do?

osprey_worker/src/osprey/engine/conftest.py
```diff
@@ -266,6 +293,16 @@ def execute_with_result_(
         action_time: Optional[datetime] = None,
     ) -> ExecutionResult:
         sources = into_sources(sources_dict)
+
+        # Ensure standard AST validators are registered before validation/execution
+        try:
+            from osprey.worker.adaptor.plugin_manager import bootstrap_ast_validators  # type: ignore
```
BinaryFiddler: Do we need `# type: ignore`?

osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_strings.py
caidanw: yeah these are a bit tricky to understand without looking at how `Scenario` is consumed. But essentially, it'll use the default value of `text` if this param is missing, otherwise it'll use this param's value to compare against the formatted string.

Basically it's saying that `∏opq` after formatting/cleaning should be `nopq`.

osprey_worker/src/osprey/engine/conftest.py
caidanw: Honestly, no clue. AI added that, I'll double check if we need it

caidanw: Good catch, wasn't needed. Updated the code.

PR #23, opened 2025-10-01
refactor: lazily initialize the bigtable client instance 

caidanw:
Summary
---

Move BigTable client initialization logic out of the `init_from_config` method and into a dedicated method `_setup_client_instance` so that we can transition the `_instance` property to be a `@property` function that lazily initializes the client.

This fixes the issue where when running in environments (like testing) that don't have GCP credentials setup locally, the application would throw the error:
```
google.auth.exceptions.DefaultCredentialsError:
Your default credentials were not found. To set up Application Default Credentials, see
https://cloud.google.com/docs/authentication/external/set-up-adc for more information.
```

There are no logical changes, with the exception that the client instance is now only initialized when it's needed.

### DIFF
--- a/osprey_worker/src/osprey/worker/lib/storage/bigtable.py
+++ b/osprey_worker/src/osprey/worker/lib/storage/bigtable.py
@@ -37,7 +37,7 @@ class BigTableClient(ABC):
 
     @property
     def _instance(self) -> Instance:
-        if not hasattr(self, '__instance'):
+        if not getattr(self, '__instance', None):
             self.__instance = self._setup_client_instance()
         return self.__instance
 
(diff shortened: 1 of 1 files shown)


osprey_worker/src/osprey/worker/lib/storage/bigtable.py
```diff
@@ -24,6 +25,22 @@
...
+
+    def _setup_client_instance(self) -> Instance:
+        client = Client(project=self._gcp_project, admin=self._admin_enabled)
+        fix_bigtable_client_if_using_emulator(client)
+
+        return client.instance(self._instance_id)
+
+    @property
+    def _instance(self) -> Instance:
+        if not hasattr(self, '__instance'):
```
EXBreder: is `if not self.__instance` insuficient here?

caidanw: That would be sufficient, but we would need to define `self.__instance` in `__init__` otherwise we will get an `AttributeError`. This way we don't need to worry about having subclasses define it, since this variable is only used as a storage mechanism.

I'm updating it to use `getattr` instead though, since my current version doesn't handle the case of someone setting this value to None.

BinaryFiddler (approved):

PR #24, opened 2025-10-02
Remove labels proto (for real this time)

EXBreder:
doing my best to remove more code.

This PR removes all label protobuf code, generated and otherwise, and replaces all instances with the "local" version. The rationale is that the labels provider, which is now provided by plugins.

This one does not provide a label provider, but that can come next. Its recommend that both a local label provider (with stuff stored in memory) and a postgres version for more (slightly) production flows.

Notable changes include:

    in the plugin bootstrap scripts, the labels UDF and labels output sink is only added if the labels provider is actually there.
    EventEffectOutputSink is just LabelOutputSink now, since it no long handles the webhooks (which also makes it a lot simpler). Its signature is altered, and calling things no longer need the engine for the sink.
    LabelStatus is moved to the engine subdirectory, making one fewer "backwards" dependency. Its also swapped to an intenum for some ease of use.

https://github.com/roostorg/osprey/pull/19 is left orphaned as it contains some trial implementations of the local provider, which is jank.

This will work as is, even if our labels provider mechanism changes.


ayubun (approved):

ayubun:
lgtm :D

PR #26, opened 2025-10-05
native support for labels

ayubun:
- refactor labels provider abstraction
  - now supports labels service OR labels provider being supplied, giving more control to the impler
- removed event effects output sink / overhauled label output sink to simply call into the provider
- fix cli, bulk label sink, and ui api for labels endpoints
  - previously, these all called into the label output sink to apply label event effects / webhooks (which have been removed -- webhooks are a leaky abstraction that migrated to osprey due to legacy code from discord's usage)
- major overhaul of in-house labels objects
  - this includes replacing all old proto references
- removed `delay_action_by` on label add / label remove
  - this allowed for the removal of `ExtendedEntityMutation` in favour of a much simpler `EntityMutation` (renamed to `EntityLabelMutation` in this PR)
- migrated service logic from labels service to the default labels provider in python
  - if someone wants to have their labels service do all of this labour, they can override the labels provider to not call this helper logic (we will need to do this for discord)
- ai slop tests to get this pr's feet on the ground

likely more but yah


osprey_worker/src/osprey/engine/stdlib/udfs/labels.py
```diff
@@ -123,7 +123,7 @@ class BatchableHasLabelArguments:
     desired_status: Optional[_SimpleStatus]
 
 
-class HasLabel(HasHelperInternal[LabelProvider], BatchableUDFBase[HasLabelArguments, bool, BatchableHasLabelArguments]):
+class HasLabel(HasHelperInternal[BaseLabelsProvider], BatchableUDFBase[HasLabelArguments, bool, BatchableHasLabelArguments]):
```
EXBreder: fwiw, I'm not a fan of the `Base` prefix - it adds verbosity when its not needed. `LabelsProvider` is already marked as an ABC, and any class that uses it doesn't need to explicitly specify that they are using a Base class, they just care that they are using something that provides labels.

ayubun: i hear u, but i think if we want to rid base, we should be doing it for all areas (i.e. UdfBase) and not just new ones. 

but i think the verbosity base provides is useful for describing ABCs at the top-level class definitions

ayubun: to match the UdfBase style tho, i think i will rename it to LabelsProviderBase lol

osprey_worker/src/osprey/worker/lib/storage/labels.py
```diff
@@ -1,27 +1,120 @@
...
+from tenacity import retry, stop_after_attempt, wait_exponential
 
 
-class LabelProvider(ExternalService[EntityT[Any], Labels], ABC):
-    def cache_ttl(self) -> Optional[timedelta]:
-        return timedelta(minutes=5)
+logger = get_logger(__name__)
+
+
+class BaseLabelsProvider(ExternalService[EntityT[Any], EntityLabels], ABC):
```
EXBreder: On abstractions - would we expect anyone to change the `apply_entity_*` functions with different service setups? I kinda expect no, though within discord we may (temporarily) extend it
 
In that case, I think it may be better to have a concrete `LabelsProvider`, but it takes in a `LabelsService` ABC, which can be implemented by someone else. Composition is usually better than inheritence, as a rule of thumb - the provider contains the business logic, where the service only deals with storage.

osprey_worker/src/osprey/worker/ui_api/osprey/views/entities.py
```diff
@@ -30,30 +30,30 @@
 def get_labels_for_entity(request_model: GetLabelsForEntityRequest) -> Any:
     require_ability_with_request(request_model, CanViewLabelsForEntity)
 
-    # TODO(ayubun): Support plug-and-play label service
-    return {
-        'labels': {},
-        'expires_at': None,
-    }
+    from ....adaptor.plugin_manager import bootstrap_labels_provider, has_labels_provider
```
EXBreder: use an absolute path here if possible

EXBreder: Q: were you running into weird circular import issues with this?

osprey_worker/src/osprey/engine/language_types/labels.py
```diff
@@ -15,11 +15,39 @@
...
+    a single attempted mutation; i.e., if an execution of the rules results in a label add and a label remove
+    of the same entity/label pair.
+    """
     ADDED = 0
     REMOVED = 1
     MANUALLY_ADDED = 2
     MANUALLY_REMOVED = 3
 
+    @property
+    def weight(self) -> int:
```
EXBreder: I was kinda thinking if we could just order the enum values accordingly instead of this - it would break as we moves stuff back into discord, but would make this a lot simpler to comprehend. wdyt

ayubun: my main fear with this is that it couples weight to the data that we store in the db; i guess i envisioned a possible world where someone added a new type here (to be annoying, like.... "LanelStatus.NEUTRAL") that would require the weights above and below to be shifted

to achieve the desired weight of the new type where the intenum value is used for weight, it would have to break any existing data stores

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -47,18 +72,134 @@ class LabelStateInner:
...
+
+        this field is a convenience value to save users time on computing the effective expiration time from the reasons.
+
+        expiration defines when future label states can be applied. if the current label state is not expired, 
+        then then upon a new label state change attempt, the current and new statuses have their weights' compared.
+        whichever has the higher weight will take precedence, and the lower weight(s) will be dropped.
+        if the weights are the *same*, then a merge of reasons is performed, which can also cause the expiration to be delayed.
+        """
+        if not reasons:
+            AssertionError(f'invariant: the label state {self} did not have any associated reasons')
```
EXBreder, line 202: Is that a true invariant? As in - why should this be an error vs just returning `None`?

ayubun: labels shouldn't be able to be modified without reasons, at least not from the examples i saw in ui 😮

perhaps i should confirm that it isn't allowed in the labels service

osprey_worker/src/osprey/engine/language_types/labels.py
EXBreder: Wait, but this is meant to be the internal representation, right? Its not out of the ordinary  for there to be a conversion step of the DB rep to the internal rep. I think we should optimize for ergonomics here

osprey_worker/src/osprey/engine/stdlib/udfs/labels.py
EXBreder: rip, though I'd prefer UdfBase to be renamed to UserDefinedFunction as well. I agree that consistency is better than moving individual stuff over to a new standard

osprey_worker/src/osprey/worker/ui_api/osprey/views/entities.py
ayubun: i wasn't, was just lazy to wait for my lsp to index and wanted the import right there lol

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -47,18 +72,134 @@ class LabelStateInner:
...
+        return bool(self.expires_at is not None and self.expires_at + timedelta(seconds=1) < datetime.now())
+
+    def _shift_current_state_to_previous_state(self) -> None:
+        if not self.reasons:
+            # to make this function idempotent, we don't want to shift an empty state to the previous state.
+            # we should always have reasons to shift
+            return
+        self.previous_states.insert(0, LabelStateInner(
+            status=self.status,
+            reasons=self.reasons.copy()
```
EXBreder: This is possibly a footgun - while we are treating reasons as immutable, its kinda not - a user of this class could take a ref to reasons and modify it later.

Probably not something _too_ bad to worry about, but we could avoid it by making the stored dict a frozen dict

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -81,26 +222,70 @@ class EntityLabelDisagreeRequest(BaseModel):
...
+    features: dict[str, str] = field(default_factory=dict)
+    expires_at: datetime | None = None
+
+    def desired_state(self) -> LabelState:
+        return LabelState(
+            status=self.status,
+            reasons={self.reason_name: self.reason()},
+        )
+
+    def reason(self) -> LabelReason:
```
EXBreder, line 289: perhaps make this a `@property`?

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -47,18 +72,134 @@ class LabelStateInner:
...
+    def update_status(self, status: LabelStatus, reasons: dict[str, LabelReason]) -> bool:
+        """
+        sets the label state to the provided status if the provided status has a higher weight or if the current
+        status is expired.
+
+        if the status is updated, the current state is shifted to previous_states and the reasons are appended.
+
+        returns true if the status is accepted and/or merged; returns false if it was dropped.
+        """
+        if self.is_expired() or self.status.weight < status.weight:
```
EXBreder: Reading the rust code again, I don't think this is right.

Weights are _not_ considered in updates of status, only Manual vs Automatic. Weights are only considered when dealing with multiple mutations from the same action, not when considering prior state.

If I read it correct, in this state, we wouldn't be able to ever add an ADDED label on top of a REMOVED label, even though that is very possible. (think about a removing then adding the same label manually because you made a mistake)

I think we need to operate on these w/ some tests to ensure we have the right contracts, this stuff is tricky and doing some TDD would probably do us some good.

osprey_worker/src/osprey/worker/lib/storage/labels.py
```diff
@@ -1,27 +1,120 @@
...
+        self, entity: EntityT[Any], mutations: Sequence[EntityLabelMutation]
+    ) -> EntityLabelMutationsResult:
+        old_labels = self.get_from_service(entity)
+        new_labels = copy.deepcopy(old_labels)
+
+        updated: list[str] = []
+        dropped: list[EntityLabelMutation] = []
+
+        # first, drop lower weight statuses and collect same-weight statuses
+        mutations_by_label_name: dict[str, list[EntityLabelMutation]] = defaultdict(list)
```
EXBreder, line 98: I think this section can be moved into a helper function just for readability that returns the mutations_by_label_name and the dropped things. This function overall is kinda hard to follow

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
ayubun: ah yeah you're totally right, that's a good catch. the logic is flawed

osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
```diff
@@ -55,13 +55,13 @@ def bootstrap_udfs() -> tuple[UDFRegistry, UDFHelpers]:
...
     # Label udfs should only be registered if the labels provider is available
-    if _has_labels_provider():
+    if has_labels_service():
         # Imports kinda circular. Imports here are to avoid that.
         from osprey.engine.stdlib.udfs.labels import HasLabel, LabelAdd, LabelRemove
 
         udfs.extend([HasLabel, LabelAdd, LabelRemove])
 
-        labels_provider = plugin_manager.hook.register_labels_provider()
+        labels_provider = LabelsProvider(plugin_manager.hook.regiser_labels_service())
```
EXBreder: This logic doesn't use the same logic as bootstrap_labels_provider(), and will fail if a provider is given instead of a service

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -1,15 +1,15 @@
...
+from typing import TYPE_CHECKING, Dict, Optional
 
-from osprey.engine.language_types.labels import LabelStatus
 from osprey.worker.lib.osprey_shared.logging import get_logger
 from osprey.worker.lib.utils.request_utils import SessionWithRetries
-from pydantic import BaseModel
 
 if TYPE_CHECKING:
-    from osprey.worker.lib.utils.flask_signing import Signer
+    pass
```
EXBreder: remove this entire if statement

osprey_worker/src/osprey/engine/language_types/labels.py
```diff
@@ -9,30 +8,12 @@
     ENTITY_LABEL_MUTATION_DIMENSION_NAME,
     ENTITY_LABEL_MUTATION_DIMENSION_VALUE,
 )
+from osprey.worker.lib.osprey_shared.labels import LabelStatus
```
EXBreder, line 11: This is kinda minor, as we don't follow it consistently, but I think we should try to keep imports originating osprey.engine of osprey.worker to a minimum (that was the reason `LabelStatus` was defined here). This is mostly to reduce circular dependencies.

I think the rough split between the two are:
engine contains abstract rule language, basics/defaults, and interfaces to define osprey
worker contains the implementations and direct execution to run osprey.

Again, since this is not consistent right now, its nbd, but something to keep in mind.

osprey_worker/src/osprey/worker/lib/storage/labels.py
```diff
@@ -1,27 +1,234 @@
...
+        removed: list[str] = []
+        updated: list[str] = []
+        new_labels = copy.deepcopy(old_labels)
+        for label_name, desired_state in desired_states_by_label_name.items():
+            if label_name not in new_labels.labels:
+                new_labels.labels[label_name] = LabelState.from_inner(desired_state)
+                added.append(label_name)
+                continue
+            current_state = new_labels.labels[label_name]
+            # tbh idk enough about python to know if this copy is necessary but better safe than sorry
```
EXBreder: not necessary, the int enum is just an int.

EXBreder (commented):
One logical change necessary, otherwise looks good. Kinda hard to follow a lot of the logic as a whole though.

osprey_worker/src/osprey/engine/language_types/labels.py
ayubun: we should move it all to engine then, eventually :o i was thinking about how we needed a refactor cuz it felt inconsistent, but it felt out of scope cuz this pr is alr chunky

osprey_worker/src/osprey/worker/adaptor/plugin_manager.py
ayubun: ahhh nice catch ! i forgot to update this

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
ayubun: this looks like a ruff autoformat :0
nice catch

EXBreder (approved):

PR #29, opened 2025-10-16
configurable execution result store

haileyok:
right now, there isn't a good way to configure or disable the execution result store. the only way i found was to use `register_execution_result_store()` in `register_plugins` but explicitly create and return a noop `ExecutionResultStore`. there's a misleading env var right now in the `docker-compose.yml` called `OSPREY_EXECUTION_RESULT_STORAGE_BACKEND` that doesn't seem to do anything as well. this results in the worker failing to process events as the requests to the minio store simply fail.

here, i took the similar route that exists for configuring the input stream
- create a new `ExecutionResultStorageBackendType` enum with Bigtable, GCS, Minio, or "plugin" as options, as well as a simple "None" option
- create a "chooser" that returns the supplied backend type
- update the sink register to use the new chooser rather than directly calling `bootstrap_execution_result_store`


osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
```diff
@@ -23,7 +25,14 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
                 kafka_producer=KafkaProducer(bootstrap_servers=bootstrap_servers, client_id=client_id),
             )
         )
-    execution_result_store = bootstrap_execution_result_store(config=config)
+
+    storage_backend_type = ExecutionResultStorageBackendType(
+        config.get_str('OSPREY_EXECUTION_RESULT_STORAGE_BACKEND', 'none')
```
haileyok: should `none` be the default, or should it be `minio` like the current behavior would result in?

osprey_worker/src/osprey/worker/lib/storage/__init__.py
```diff
@@ -0,0 +1,30 @@
+from enum import StrEnum, auto
+
+
+class ExecutionResultStorageBackendType(StrEnum):
```
haileyok, line 11: afaict, the ones i added here are the only ones that exist

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -482,9 +484,16 @@ def get_many(
...
-    storage_backend = bootstrap_execution_result_store(config)
+
+    storage_backend_type = ExecutionResultStorageBackendType(
+        config.get_str('OSPREY_EXECUTION_RESULT_STORAGE_BACKEND', 'none')
+    )
+    storage_backend = get_rules_execution_result_storage_backend(backend_type=storage_backend_type)
+
+    if storage_backend is None:
+        raise AssertionError('No storage backend registered')
+
```
haileyok, line 513: afaict, it makes sense to raise here. the codepaths i see calling `bootstrap_execution_result_storage_service` are all in the ui api or in one of the cli tools, and they probably should be raising an error when there isn't a configured backend.

osprey_worker/src/osprey/worker/lib/storage/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    MINIO = auto()
+    """
+    Minio execution result store
+    """
+
+    PLUGIN = auto()
+    """
+    Execution result store that is sdefined via register_execution_result_store
```
ayubun:
```suggestion
    Execution result store that is defined via register_execution_result_store
```

osprey_worker/src/osprey/worker/lib/storage/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    NONE = auto()
+    """
+    Disable execution results from being stored
```
ayubun: it may be helpful to tell users what the consequences of using this are
```suggestion
    Disable execution results from being stored. This may cause certain elements of Osprey to break, such as the events stream and individual event details in the UI
```

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -482,9 +483,17 @@ def get_many(
...
     """Create an ExecutionResultStorageService with the configured storage backend."""
-    from osprey.worker.adaptor.plugin_manager import bootstrap_execution_result_store
+    from osprey.worker._stdlibplugin.execution_result_store_chooser import get_rules_execution_result_storage_backend
     from osprey.worker.lib.singletons import CONFIG
 
     config = CONFIG.instance()
-    storage_backend = bootstrap_execution_result_store(config)
+
+    storage_backend_type = ExecutionResultStorageBackendType(
+        config.get_str('OSPREY_EXECUTION_RESULT_STORAGE_BACKEND', 'none')
```
ayubun:
```suggestion
        config.get_str('OSPREY_EXECUTION_RESULT_STORAGE_BACKEND', 'none').upper()
```
do python enums act case insensitive ? if not, all the types seem to only accept upper so we might as well `upper()` here for robustness. if so, ignore me 😆

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
ayubun: i think none, because minio is extra setup cost;

with that being said, i think we can have minio setup in the example plugins (?) we talked about moving it there during our sync today

osprey_worker/src/osprey/worker/lib/storage/__init__.py
haileyok: neovim no spellcheck moment

osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
haileyok: so ig `auto()` [makes the strings lower](https://docs.python.org/3/library/enum.html#enum.StrEnum) so i think we atually want `.lower()`

haileyok:
```suggestion
        config.get_str('OSPREY_EXECUTION_RESULT_STORAGE_BACKEND', 'none').lower()
```

osprey_worker/src/osprey/worker/_stdlibplugin/sink_register.py
haileyok: ya for now i left it here https://github.com/roostorg/osprey/blob/701f3d589f6931ad8340ba7da60acfb4b092ed4c/osprey_worker/src/osprey/worker/_stdlibplugin/storage_register.py#L7

seems like a fine thing tho to move to examples, i can clean that up if ppl have opinions on what to do with it

EXBreder (approved):

haileyok:
fyi @cmttt idk if this affects how you're doing things but just flagging incase it does

PR #31, opened 2025-10-17
add postgres-backed labels service example & unify docker compose names

ayubun:
---

- added a postgres-backed labels service implementation in the example plugins, which connects to the postgres db that we already stand up
- fixed the ui for label mutations https://github.com/roostorg/osprey/pull/39
- reworked the labels service / labels provider abstractions a bit for ergonomics
- unified the naming scheme that we use for docker compose
  - we had like. . . underscore-separators and dash-separators . . and it just didn't make sense. so i moved every container name / host name / compose service name to dash separators, and of course file directories / file names remain underscore separated
  - some services had a `hostname` field, while some did not. i gave every service a `hostname` that matches the `container_name`, because.. why not ?
- fixed local dev
  - hostname not resolving issues when trying to contact osprey-ui-api:5004 kept plaguing me when port forwarding the ui to my local machine, so i updated the react api url to point to localhost:5004 and all became well
  - druid was trying to reach zookeeper at `zookeeper`, but i updated the host to `druid-zookeeper` a short bit ago
  - i modified snowflake to use port `8088` because `8080` i believe is used by druid router to display the druid ui; either way, 8088 is more obscure/less likely to conflict, so why not
  - BanUser example UDF had an invalid format to be properly recognized as an effect (it needed to return `EffectBase` directly, so that it can be used in-line with other effects in a single WhenRules)

---


example_plugins/src/register_plugins.py
```diff
@@ -16,3 +18,9 @@ def register_udfs() -> Sequence[Type[UDFBase[Any, Any]]]:
 @hookimpl_osprey
 def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
     return [StdoutOutputSink(log_sampler=None)]
+
+
+@hookimpl_osprey
+def register_labels_service(config: Config) -> LabelsServiceBase:
+    """Register a PostgreSQL-backed labels service."""
+    return PostgresLabelsService(database='osprey_db')
```
ayubun: doesnt need to specify this since default value is osprey_db

example_plugins/src/services/labels_service.py
```diff
@@ -0,0 +1,115 @@
...
+        with scoped_session(commit=False, database=self.database) as session:
+            try:
+                # Use SELECT FOR UPDATE to acquire a row-level lock
+                stmt = select(EntityLabelsModel).where(EntityLabelsModel.entity_key == entity_key).with_for_update()
+                result = session.execute(stmt).scalars().first()
+
+                if result is None:
+                    labels = EntityLabels()
+                else:
+                    labels = EntityLabels.deserialize(result)
```
ayubun:
```suggestion
                    labels = EntityLabels.deserialize(result.labels)
```

example_plugins/src/services/labels_service.py
```diff
@@ -0,0 +1,115 @@
...
+
+        with scoped_session(database=self.database) as session:
+            stmt = select(EntityLabelsModel).where(EntityLabelsModel.entity_key == entity_key)
+            result = session.execute(stmt).scalars().first()
+
+            if result is None:
+                logger.debug(f'No labels found for entity {entity_key}')
+                return EntityLabels()
+
+            labels = EntityLabels.deserialize(result)
```
ayubun:
```suggestion
            labels = EntityLabels.deserialize(result.labels)
```

osprey_worker/src/osprey/worker/lib/storage/labels.py
```diff
@@ -71,10 +70,16 @@ def batch_read_labels(self, entities: Sequence[EntityT[Any]]) -> Sequence[Result
...
-    def get_labels_atomically(self, entity: EntityT[Any]) -> Generator[EntityLabels, None, None]:
+    def read_modify_write_labels_atomically(self, entity: EntityT[Any]) -> Generator[EntityLabels, None, None]:
         """
-        Context manager for atomic read-modify-write operations.
-        Implementations should ensure the entity key is locked/in a transaction.
+        Context manager for atomic read-modify-write operations. This generator should yield EntityLabels upon reading
+        and should write the EntityLabels post-yield.
+
+        IMPORTANT: Implementations should ensure the entity key is locked/in a transaction so that other read-modify-write
+                   calls must wait.
```
ayubun:
```suggestion
                   calls (even across multiple workers) must wait.
```

example_plugins/src/services/labels_service.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, database: str = 'osprey_db'):
+        """
+        Initialize the PostgreSQL labels service.
+
+        Args:
+            database: The database name to use. Defaults to 'osprey_db'.
+        """
+        self.database = database
+        # Initialize the database connection from config
+        init_from_config(database)
```
EXBreder: Heavily prefer to _not_ do side effect inducing things within the __init__ method - its not something expected as a caller. 

Instead, I think we can add in an initialize method to the service base, and call that when initialization should actually occur.

example_plugins/src/services/labels_service.py
```diff
@@ -0,0 +1,115 @@
...
+                upsert_stmt = upsert_stmt.on_conflict_do_update(index_elements=['entity_key'], set_=labels_dict)
+                session.execute(upsert_stmt)
+
+                session.commit()
+                logger.debug(f'Committed atomic read-modify-write for entity {entity_key}')
+
+            except Exception:
+                session.rollback()
+                logger.error(f'Rolled back atomic read-modify-write for entity {entity_key}')
+                raise
```
EXBreder, line 125: Should raise the exception received instead of blank (unless python does that automagically?)

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
```diff
@@ -98,12 +98,39 @@ class LabelReason:
...
 
     if a given label state has multiple label reasons, all reasons would need to expire before the status/state
-    is considered expired, too. 
+    is considered expired, too.
     """
 
     def is_expired(self) -> bool:
         return bool(self.expires_at is not None and self.expires_at + timedelta(seconds=5) < datetime.now())
 
+    def serialize(self) -> dict[str, Any]:
```
EXBreder, line 115: Is there a reason to do this vs using `dataclasses.asdict()`?

example_plugins/src/services/labels_service.py
ayubun: ack, i made an initialize() method on service and provider, lmk what u think

example_plugins/src/services/labels_service.py
ayubun: https://stackoverflow.com/questions/25632147/raise-at-the-end-of-a-python-function-outside-try-or-except-block

osprey_worker/src/osprey/worker/lib/osprey_shared/labels.py
ayubun: more control; namely, datetime objects cannot be converted via `json.dumps()`, which is important for database serialization/deserialization. thus, we convert them to iso format here, and it also allows us to add some magic to ensure that all our datetimes are always timezone-aware, which i added in this last iteration

init-minio-bucket.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-mc alias set myminio http://minio:9000 minioadmin minioadmin123
+mc alias set myminio http://localhost:9000 minioadmin minioadmin123
```
ayubun: this might be necessary to support ui on local machine, but it's hard to say because i've yet to get test data working

osprey_worker/src/osprey/worker/lib/cli.py
```diff
@@ -325,69 +323,15 @@ def bulk_apply_label(
...
+        _ = provider.apply_entity_label_mutations(
             entity=EntityT(type=entity_type, id=entity_id),
             mutations=[mutation],
         )
         progress_tracker.increment()
 
     print(f'Bulk labelling complete! Total labels applied: {progress_tracker.total_actions}')
-
-
-class IpAddress(ParamType):
```
ayubun, line 339: this is unused everywhere

osprey_worker/src/osprey/worker/lib/cli.py
```diff
@@ -325,69 +323,15 @@ def bulk_apply_label(
...
-        ipaddress.IPv4Address(value)
-        return value
-
-
-@cli.command()
-@click.option('--from-sub-gcp-project', required=True, help='Source GCP project for the subscription')
-@click.option('--from-subscription', required=True, help='Subscription ID to read from')
-@click.option('--target-destination-gcp-topic', required=True, help='Target GCP project for the topic')
-@click.option('--destination-topic', required=True, help='Destination topic to publish to')
-def restream_subscription(
```
ayubun, line 351: discord only

osprey_worker/src/osprey/worker/lib/storage/bigtable.py
```diff
@@ -89,63 +89,4 @@ def table(self, table_name: str) -> Table:
         return t
 
 
-class DataServicesBigTable(BigTableClient):
```
ayubun, line 92: discord only

osprey_worker/src/osprey/worker/lib/storage/bigtable.py
```diff
@@ -89,63 +89,4 @@ def table(self, table_name: str) -> Table:
...
-        t = self._instance.table(table_name)
-        pin_override(
-            t,
-            service='osprey-bigtable-client',
-            tags={'bigtable_instance': self._instance.instance_id, 'table_id': t.table_id},
-        )
-        return t
-
-
-class DataStreamBigTable(BigTableClient):
```
ayubun, line 120: discord only

init-minio-bucket.sh
ayubun: i changed this back to `minio:`; the ui api should handle this networking, so it should be able to use the docker network just fine

docker-compose.yaml
```diff
@@ -221,17 +229,19 @@ services:
...
-    container_name: kafka_test_data
+    hostname: kafka-test-data-producer
+    container_name: kafka-test-data-producer
     depends_on:
       kafka:
         condition: service_healthy
       kafka-topic-creator:
         condition: service_completed_successfully
     profiles:
       - test_data
```
EXBreder, line 243: why both

osprey_worker/src/osprey/worker/sinks/sink/bulk_label_sink.py
```diff
@@ -424,7 +422,10 @@ def rollback_task_effects(
         rows_rolled_back = 0
 
         feature_name_to_entity_type_mapping = engine.get_feature_name_to_entity_type_mapping()
-        labels_provider = bootstrap_labels_provider()
+        labels_provider = LABELS_PROVIDER.instance()
+        assert labels_provider is not None, (
```
EXBreder: nit: NotImplementedError would be better here than a plain assert

osprey_worker/src/osprey/worker/lib/cli.py
```diff
@@ -325,69 +323,15 @@ def bulk_apply_label(
...
 
     progress_tracker: CliCommandProgressTracker = CliCommandProgressTracker(total_actions=len(entity_ids))
-    provider = bootstrap_labels_provider()
+    provider = LABELS_PROVIDER.instance()
+    assert provider is not None, (
+        'this code cannot be used because no labels service / provider is supplied for this osprey instance'
+    )
     for entity_id in entity_ids:
-        provider.apply_entity_label_mutations(
+        _ = provider.apply_entity_label_mutations(
```
EXBreder, line 331: huh, why is the `_ =` necessary?

osprey_worker/src/osprey/worker/lib/cli.py
```diff
@@ -260,10 +253,13 @@ def apply_label(
             description=description or 'Manually changed from the command line for debugging.',
         )
 
-    result = bootstrap_labels_provider().apply_entity_label_mutations(
-        entity=EntityT(type=entity_type, id=entity_id), mutations=[mutation]
+    provider = LABELS_PROVIDER.instance()
```
EXBreder, line 256: Myabe have the instance() method raise a notimplementederror, instead of having an assert everywhere?

osprey_worker/src/osprey/worker/ui_api/osprey/views/entities.py
```diff
@@ -94,4 +92,12 @@ def manual_entity_mutation(request_model: ManualEntityLabelMutationRequest) -> A
 
     result = labels_provider.apply_entity_label_mutations(request_model.entity, mutations)
 
-    return result
+    return {
```
EXBreder: Can this and the EntityLabelMutationResult.serialize method be shared? Its a slightly different format in both cases, but its weird that its different.

example_rules/rules/post_contains_hello.sml
```diff
@@ -15,5 +15,8 @@ ContainsHello = Rule(
 
 WhenRules(
   rules_any=[ContainsHello],
-  then=[BanUser(entity=UserId, comment='User said "hello"')],
+  then=[
+    BanUser(entity=UserId, comment='User said "hello"'),
+    LabelAdd(entity=UserId, label='meow'),
```
EXBreder, line 20: meow

EXBreder (approved):

docker-compose.yaml
ayubun: it just felt inconsistent with the naming schemes to not support a dash version too

osprey_worker/src/osprey/worker/lib/cli.py
ayubun: just a basedpyright warning that the output is unused ╮ (. ❛ ᴗ ❛.) ╭

osprey_worker/src/osprey/worker/lib/cli.py
ayubun: well, most areas that try to use the labels provider should actually fail gracefully; i considered having it throw the exception but it felt less clean, especially since mypy wont guarantee the exception is handled by future implementers but it will guarantee `None` is addressed

osprey_worker/src/osprey/worker/ui_api/osprey/views/entities.py
ayubun: it can be shared, yea. it's different because the ui expects this weird format and i'm not really pressed to fix it

osprey_worker/src/osprey/worker/lib/cli.py
EXBreder: very fair

EXBreder (approved):

PR #36, opened 2025-10-22
run bigtable rpcs in their own thread

EXBreder:
100% slop, but this matches with my understanding of the gevent <> BT issues.


ayubun (approved):
sure i guess 😆

haileyok (approved):

PR #37, opened 2025-10-22
use flask-cors to fix preflight cors errors

haileyok:


ayubun (approved):
i trust u with my life

PR #41, opened 2025-10-28
fix(tests): stabilize worker test suite and improve test infrastructure

caidanw:
### Summary

- Comprehensive overhaul of the worker test infrastructure to resolve reliability issues
- Separated test and production Docker compose configurations for better isolation
- Fixed multiple database-related test failures including PostgreSQL enum conflicts and SQLAlchemy model registration issues

#### Test Infrastructure & Docker Configuration

- Added dedicated test Docker configuration (`docker-compose.test.yaml`) that extends the main compose file with test-specific overrides
- Implemented automated test script (`run-tests.sh`) with proper service orchestration, volume cleanup, and dependency management
- Added comprehensive healthchecks for postgres, etcd, and other services to ensure proper startup ordering
- Moved etcd service to test-only configuration to reduce production complexity

#### Database & Storage Fixes

- Resolved PostgreSQL enum naming conflicts by using distinct enum names (`job_status` vs `task_status`) in bulk action models
- Fixed SQLAlchemy model registration by ensuring all models are imported before table creation in `postgres.py`
- Updated timezone handling for DateTime fields in `TemporaryAbilityToken` to use timezone-aware columns
- Fixed enum type creation for bulk label task status to prevent database errors

#### Test Configuration & Fixtures

- Made `Config.configure` idempotent during test sessions to prevent "already been bound" errors when multiple fixtures call configuration
- Aligned database configuration to use `osprey_db` key consistently across tests, compose files, and CLI
- Moved configuration setup to `worker/conftest.py` for better test isolation and removed redundant fixtures
- Updated test expectations for processed rows, query counts, and other metrics to match actual system behavior

#### Code Quality & Maintenance

- Removed deprecated `EntityLabelWebhook` tests as the feature is no longer supported
- Updated type annotations for `LabelsProvider` initialization to use forward references
- Fixed import order issues and removed unused parameters from test data
- Updated pre-commit configuration by removing problematic `check-yaml` hook (was conflicting with valid `docker-compose.yaml` syntax, specifically `volumes: !reset []`)
- Skipped manual-only tests that were causing CI failures

#### Service Dependencies & Environment

- Fixed `entrypoint.sh` to properly run tests with correct working directory and use uv to run pytest execution
- Updated `SNOWFLAKE_API_ENDPOINT` configuration for test environment compatibility
- Enhanced Docker service dependencies with proper wait conditions and restart policies
- Improved volume handling with reset directives for test containers to prevent state leakage


.pre-commit-config.yaml
```diff
@@ -11,7 +11,6 @@ repos:
         exclude: .*/tests?/.*\.txt
       - id: end-of-file-fixer
         exclude: .*/tests?/.*\.txt
-      - id: check-yaml
```
ayubun, line 14: what does removing this do / why (?)

osprey_worker/src/osprey/worker/lib/config/__init__.py
```diff
@@ -241,7 +241,7 @@ def debug(self) -> bool:
         return self.get_bool('DEBUG', False)
 
     def __getitem__(self, item: str) -> Any:
-        return self._config_dict[item]
+        return self._config_dict.get(item)
```
ayubun: does this not return a `Optional`? why do we override this?

ayubun (approved):

.pre-commit-config.yaml
caidanw: I put a not in the PR description about this:
> Updated pre-commit configuration by removing problematic check-yaml hook (was conflicting with valid docker-compose.yaml syntax, specifically volumes: !reset [])

It's because the check-yaml hook doesn't cover valid syntax for docker compose files. We could exclude just these files but honestly that check isn't super valuable, since it doesn't recognize app specific configs.

osprey_worker/src/osprey/worker/lib/config/__init__.py
caidanw: I should probably update the return type of the function.
This function throws an error if the key is not found in the dictionary.
Is throwing an error the desired functionality here?

caidanw: I've reverted the change

ayubun (approved):

PR #43, opened 2025-10-30
User Personas

andrewmchang:
Adding user research to Osprey documentation


julietshen:
also another nit: could you add text details for the images too? just in case anyone's using a screen reader, it also helps break info out of images generally

julietshen (approved):
LGTM!

PR #44, opened 2025-10-30
fix label tests

ayubun:
this pr addresses both `test_labels.py` files (yes there's two)

it does not address the following database issues that were recurrent during the `./run-tests.sh` iterations:
Screenshot 2025-10-30 at 1 44 04 PM


caidanw (approved):
LGTM

haileyok (approved):
nice

PR #45, opened 2025-10-31
ui: fallback text copying for insecure hosts

haileyok:
## Problem

The async pubsub publisher was observed dropping 1 message in ~23,200 publishes (99.987% success rate) due to a `concurrent.futures.TimeoutError` on `future.result(timeout=5)`. No retry was attempted — the message was silently lost.

## Solution

Replaces the flat `_sync_flush` loop with a `_publish_one_with_retry` helper that retries each message independently with bounded exponential backoff + full jitter.

### Retry policy

- **Max attempts:** 3 (1 initial + 2 retries)
- **Retryable errors:** `concurrent.futures.TimeoutError` and gRPC `UNAVAILABLE`, `DEADLINE_EXCEEDED`, `RESOURCE_EXHAUSTED`, `ABORTED`, `INTERNAL`
- **Non-retryable errors:** `NOT_FOUND`, `PERMISSION_DENIED`, `INVALID_ARGUMENT`, and anything else — fail immediately, no delay
- **Backoff:** base 0.5s × 2^(attempt−1), capped at 10s, full jitter applied via `random.uniform(0, delay)`

### No silent loss

Every message either succeeds or emits `permanent_failure`. The old code emitted `failure` and moved on; now it also tags the terminal outcome explicitly so dashboards can track the residual rate.

## New metrics

| Metric | Tags | When |
|--------|------|------|
| `async_pubsub_publisher.publish.retried` | `attempt:N` | Each retry attempt before success or exhaustion |
| `async_pubsub_publisher.publish.permanent_failure` | `error:<ClassName>` | Message exhausted retries or hit non-retryable error |

Existing `publish.attempt`, `publish.success`, and `publish.failure` are preserved unchanged. `failure` fires on every failed attempt (including ones that will be retried).

## Note

This complements the `publish.attempt/success/failure` metrics patch deployed today (now on main as #44).

## Test plan

- [ ] `test_single_attempt_success` — happy path unchanged
- [ ] `test_transient_error_then_success` — TimeoutError on attempt 1, success on attempt 2
- [ ] `test_transient_error_exhausts_retries` — 3 attempts all fail, permanent_failure metric fires
- [ ] `test_non_retryable_error_no_retry` — NOT_FOUND skips retry, goes straight to permanent_failure
- [ ] `test_sync_flush_publishes_all_messages` — batch of 3 all succeed
- [ ] `test_sync_flush_one_failure_does_not_block_others` — failed message doesn't prevent rest of batch


osprey_ui/src/lib/copyText.ts
```diff
@@ -0,0 +1,20 @@
...
+// we don't want to break things when a user tries to copy text somewhere. here we use a
+// fallback when either that api is not available or `secureContext` is false
+// see: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard
+export async function copyText(text: string): Promise<void> {
+  if (navigator.clipboard && window.isSecureContext) {
+    return navigator.clipboard.writeText(text);
+  } else {
+    // create a textarea that gets appended to the body, then focus+copy text
+    const body = document.body;
+    const textarea = document.createElement('textarea');
```
ryanprior, line 12:
```suggestion
    const textarea = document.createElement('textarea');
    textarea.setAttribute('aria-hidden', 'true');
```
Detail: you can set `aria-hidden` before appending to the document body to ensure it doesn't proc assistive tech.

osprey_ui/src/lib/copyText.ts
```diff
@@ -0,0 +1,20 @@
...
+    body.appendChild(textarea);
+    textarea.value = text;
+    textarea.focus();
+    textarea.select();
+    document.execCommand('copy');
+    textarea.remove();
```
ryanprior:
```suggestion
    try {
      document.execCommand('copy');
    }
    catch (err) {
      console.log(`copyText: fallback failed: ${err}`);
    }
    finally {
      textarea.remove();
    }
```

Since the `copy` command is [formally deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand), consider wrapping it in `try` to ensure that the cleanup textarea removal fires even if the browser vendors decide to break this down the road. (Or with the various new browsers in development, who knows if they'll choose to implement deprecated functionality or not.)

ryanprior (commented):
I like this fallback idea :)

osprey_ui/src/lib/copyText.ts
haileyok: cool!

osprey_ui/src/lib/copyText.ts
haileyok: that makes sense! i changed this to actually log an error instead of a generic info log so that it's more obvious (there might be an alert system somewhere in this codebase that im unaware of that could get used too...)

BinaryFiddler (approved):

cassidyjames (approved):
This seems reasonable; @ThisIsMissEm I filed your comment as #82 so we don't lose it, because that also makes sense to me!

PR #46, opened 2025-10-31
ci: setup integration tests

caidanw:
Summary
---

Re-use the `run-tests.sh` script in a new `integration-tests.yml` workflow.
Both build the docker images and run the tests to help prevent regressions.

**IMPORTANT:** do not merge this PR until all tests are passing, otherwise it will prevent anyone from merging their PRs.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,41 @@
+name: Integration Tests
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    branches: [main]
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Make run-tests.sh executable
+        run: chmod +x run-tests.sh
+
+      - name: Create temp directory for test results
+        run: mkdir -p /tmp/test-results
+
+      - name: Run integration tests
+        run: ./run-tests.sh --junitxml=/tmp/test-results/junit-pytest.xml
+
+      - name: Clean up Docker containers
+        if: always()
+        run: docker compose -f docker-compose.yaml -f docker-compose.test.yaml --profile test down --volumes
+
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        if: always()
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    branches: [main]
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
```
EXBreder: nit: pick an aribitrary ubuntu image, but peg it. This always has a chance of updating down the line w/o code change, causing a problem, and being hard to debug

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Make run-tests.sh executable
+        run: chmod +x run-tests.sh
```
EXBreder: you should be able to just chmod this directly in the repo, w/o requiring a build step to do it

caidanw: It is already, but it's good practice. I'll remove it though

caidanw: Fixed

.github/workflows/integration-tests.yml
caidanw: Fixed

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Pull and build Docker images
+        run: |
+          docker compose -f docker-compose.yaml -f docker-compose.test.yaml --profile test pull
+          docker compose -f docker-compose.yaml -f docker-compose.test.yaml --profile test build
+
+      - name: Run integration tests
+        run: ./run-tests.sh --junitxml=/tmp/test-results/junit-pytest.xml
+
+      - name: Clean up Docker containers
```
EXBreder: Is cleanup actually necessary? I'd assume we get a fresh environment each time.

EXBreder (approved):

.github/workflows/integration-tests.yml
caidanw: Not necessarily, but it's good practice. We can remove it if you'd like

caidanw: removed

PR #48, opened 2025-11-03
cli: small tweaks to get `push-rules` working

haileyok:
Trying to get CI tests and etcd rule pushing working, ran into a couple of problems/quirks.

- No clear way to run any of the commands inside of `cli.py` (might actually want to add this to `entrypoint.sh` too? :woman_shrugging:)
- When running `cli.py`, being yelled at about a missing `DESCRIPTOR` attribute because of using `EnumChoicePb2`
- Not bootstrapping AST validators before copying the validator registry that gets passed to `validate_sources`

Cleaning up all of those here. Not 100% sure why the decision for `EnumChoicePb2` was there, so that in particularly might need to get changed back/tweaked

Can test these changes by running `uv run osprey-cli push-rules example_rules/ --dry-run` after pulling this branch.

Also eventually worth documenting that you can do rules validation locally like this, cc @julietshen (or...whoever is going to write docs... :sweat_smile:)

### DIFF
--- a/osprey_worker/pyproject.toml
+++ b/osprey_worker/pyproject.toml
@@ -8,5 +8,7 @@ dependencies = [
     "flask-cors>=6.0.1",
     "osprey_rpc",
 ]
+[project.scripts]
+osprey-cli = "osprey.worker.lib.cli:cli"
 [project.entry-points."osprey_plugin"]
 stdlib = "osprey.worker._stdlibplugin"
--- a/osprey_worker/src/osprey/worker/lib/cli.py
+++ b/osprey_worker/src/osprey/worker/lib/cli.py
@@ -30,7 +30,7 @@ from osprey.worker.lib.storage import (  # noqa: E402
     labels,
     stored_execution_result,
 )
-from osprey.worker.lib.utils.click_utils import EnumChoicePb2  # noqa: E402
+from osprey.worker.lib.utils.click_utils import EnumChoice  # noqa: E402
 
 
 @click.group()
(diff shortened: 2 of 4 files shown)


osprey_worker/src/osprey/worker/lib/cli.py
```diff
@@ -204,7 +204,7 @@ def get_lines_from_file_as_set(file_path: str) -> Set[str]:
 @click.argument('entity_type')
 @click.argument('entity_id')
 @click.argument('label_name')
-@click.argument('label_status', type=EnumChoicePb2(LabelStatus))
+@click.argument('label_status', type=EnumChoice(LabelStatus))
```
haileyok, line 207: don't really know the context on why `EnumChoicePb2` was used here, i'm very likely missing something. feel free to rec a better way of fixing this if this is breaking

osprey_worker/pyproject.toml
```diff
@@ -8,5 +8,7 @@ dependencies = [
     "flask-cors>=6.0.1",
     "osprey_rpc",
 ]
+[project.scripts]
+osprey-cli = "osprey.worker.lib.cli:cli"
```
haileyok, line 12: a quick google suggested doing this rather than adding the whole `if __name__ == '__main__`' to `cli.py`. also feels nicer to me but could do the other/both

osprey_worker/src/osprey/worker/lib/cli.py
ayubun: this is because label status used to be a pb2 protobuf, but this has since changed; enumchoicepb2 seems like a remnant / bug

ayubun: could you actually delete this `EnumChoicePb2` class in click utils as a part of this pr ? it's unneeded 👀

ayubun (approved):

ayubun (approved):

PR #50, opened 2025-11-08
feat: add rust build workflow

caidanw:
Summary
---

Added an example app in rust under the `osprey_coordinator` directory.
Then updated the Code Quality Checks Github workflow for checking rust formatting, linting, building, and testing.

Also updated the Code Quality Checks workflow for triggering on push to the main branch as well.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Build Rust project
         run: |
-          apt-get install protobuf-compiler
+          sudo apt-get install protobuf-compiler
           cargo build --verbose
         working-directory: osprey_coordinator
 


.github/workflows/code-quality.yml
```diff
@@ -59,3 +61,45 @@ jobs:
       - name: Run prettier check
         run: npm run format:check
         working-directory: osprey_ui
+
+  rust-quality:
```
BinaryFiddler, line 65: Let's maybe disable this for for now, I don't expect the coordinator to work for a bit, I don't want it to be a blocker for other PRs.

caidanw: Disable the whole workflow or just part of it? I think it's reasonable to disable the linting and testing for now, but it's also reasonable to expect that main branch should always have a valid compiling version. Thoughts?

caidanw: Fixed!

PR #57, opened 2025-11-14
Fix osprey coordinator compilation

BinaryFiddler:
This does not run yet, but the code can now compile.

- Fixed all the pathings related to the proto files
- Added some missing proto constructs
- Renamed some smite references to osprey

Next steps:
- Remove deprecated Label service related code
- Compiler shows a fair amount of warnings about unused code, remove them would be a nice next step

### DIFF
--- a/osprey_coordinator/src/proto/mod.rs
+++ b/osprey_coordinator/src/proto/mod.rs
@@ -1,66 +1,50 @@
 #![allow(unused_imports)]
 
-pub mod discord_smite_rpc {
-    pub mod actions {
-        pub mod v1 {
-            include!(concat!(env!("OUT_DIR"), "/discord_smite_rpc.actions.v1.rs"));
-        }
-    }
-
-    pub mod common {
-        pub mod v1 {
-            include!(concat!(env!("OUT_DIR"), "/discord_smite_rpc.common.v1.rs"));
-        }
-    }
-
-    pub mod labels {
-        pub mod v1 {
-            include!(concat!(env!("OUT_DIR"), "/discord_smite_rpc.labels.v1.rs"));
-        }
-    }
-
-    pub mod smite_coordinator {
-        pub mod bidirectional_stream {
+pub mod osprey {
+    pub mod rpc {
+        pub mod actions {
             pub mod v1 {
-                include!(concat!(
-                    env!("OUT_DIR"),
-                    "/discord_smite_rpc.smite_coordinator.bidirectional_stream.v1.rs"
-                ));
+                include!(concat!(env!("OUT_DIR"), "/osprey.rpc.actions.v1.rs"));
+            }
+        }
+        pub mod common {
+            pub mod v1 {
+                include!(concat!(env!("OUT_DIR"), "/osprey.rpc.common.v1.rs"));
             }
         }
 
-        pub mod sync_action {
+        pub mod labels {
             pub mod v1 {
-                include!(concat!(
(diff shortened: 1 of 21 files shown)


proto/osprey/rpc/labels/v1/service.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+package osprey.rpc.labels.v1;
+
+// Mock proto file for deprecated labels functionality
+// These are marked as DEPRECATED in sync_action service
+// and will be removed once labels are replaced with verdicts
+
+message EntityKey {
```
BinaryFiddler, line 9: These seems to be the bare minimum to make the build pass without touching the Rust code. I'm add it for now just to make the build work. It will be removed in a separate PR when I remove calls to label service all together.

proto/osprey/rpc/actions/v1/user.proto
```diff
@@ -31,5 +30,11 @@ message User {
   google.protobuf.StringValue global_name = 21;
   // Used by guild automation GUILD_JOINED event
   google.protobuf.StringValue hub_email = 22;
-  discord_protos.users.v1.UserAvatarDecoration avatar_decoration_data = 23;
+  UserAvatarDecoration avatar_decoration_data = 23;
+}
+
+message UserAvatarDecoration {
```
BinaryFiddler: This is inlined to maintain backward compatibility back into discord land

proto/osprey/rpc/actions/v1/action.proto
```diff
@@ -141,16 +141,27 @@ message BatchMLCluster {
...
   google.protobuf.Timestamp approx_last_used_time = 5;
-  discord_protos.discord_authentication.v1.UserSessionData.ClientInfo client_info = 6;
+  ClientInfo client_info = 6;
   // Previous client info of the session if a change was detected
-  discord_protos.discord_authentication.v1.UserSessionData.ClientInfo prev_client_info = 7;
+  ClientInfo prev_client_info = 7;
   google.protobuf.Timestamp prev_approx_last_used_time = 8;
 }
 
+enum SessionVersion {
```
BinaryFiddler: Both the `SessionVersion` and `ClientInfo` are inlined to maintain backward compatibility back into discord land

osprey_coordinator/src/proto/mod.rs
```diff
@@ -1,66 +1,50 @@
...
-                include!(concat!(
-                    env!("OUT_DIR"),
-                    "/discord_smite_rpc.smite_coordinator.bidirectional_stream.v1.rs"
-                ));
+                include!(concat!(env!("OUT_DIR"), "/osprey.rpc.common.v1.rs"));
             }
         }
 
-        pub mod sync_action {
+        pub mod labels {
```
BinaryFiddler, line 16: This will be removed when I remove the label service calls in `sync_action.rs`. I don't want to touch the code logic in the same PR.

proto/osprey/rpc/actions/v1/user.proto
EXBreder: Is there a way to keep the backwards compatability without defining these directly? i.e., keep 23 reserved without pointing at anything.

proto/osprey/rpc/actions/v1/action.proto
EXBreder: reserve instead of defining

PR #60, opened 2025-11-18
Remove label service proto code

BinaryFiddler:
Removing label service proto and related logic in osprey coordinator in favor of verdict


BinaryFiddler:
superceded by https://github.com/roostorg/osprey/pull/61/

PR #63, opened 2025-11-18
Remove references of Discord in osprey coordinator

cmttt:


osprey_coordinator/etcd_config_derive/src/disconfigs.rs
```diff
@@ -29,7 +29,7 @@ impl Disconfig {
             .meta;
 
         let config_name = ConfigName::from_meta(config_name_meta)?.name.value();
-        let key_root = format!("/discord_configs/{config_name}/v1/");
+        let key_root = format!("/osprey_configs/{config_name}/v1/");
```
cmttt, line 32: Disconfigs is a way Discord handles updating configs live, it's not used by smite-coordinator, so I think we should be able to get rid of these files @BinaryFiddler

osprey_coordinator/src/pigeon/health.rs
```diff
@@ -9,7 +9,7 @@ use tonic_health::pb::{
     HealthCheckResponse,
 };
 
-const HEALTH_CHECK_ALL_PSEUDO_SERVICE: &str = "discord.v1.HealthCheckAll";
+const HEALTH_CHECK_ALL_PSEUDO_SERVICE: &str = "osprey.v1.HealthCheckAll";
```
cmttt, line 12: This name only matters when doing manual health checks.

osprey_coordinator/src/pubsub.rs
```diff
@@ -205,14 +205,12 @@ pub async fn start_pubsub_subscriber(
...
-        PubSubSubscription::new("discord-dev", "rules-sink")
-    } else {
+    let subscription_name = {
         let project_id = std::env::var("PUBSUB_SUBSCRIPTION_PROJECT_ID")
-            .unwrap_or("discord-anti-abuse-prd".to_string());
+            .unwrap_or("osprey-dev".to_string());
 
         let subscription_id = std::env::var("PUBSUB_SUBSCRIPTION_ID")
-            .unwrap_or("osprey-coordinator-test".to_string());
+            .unwrap_or("osprey-coordinator-actions".to_string());
```
cmttt, line 213: `PUBSUB_EMULATOR_HOST` is just a way to signify whether we are in DEV environment or not. Truthfully, both the project and subscription should both just be handled by separate env variables.

BinaryFiddler (approved):

PR #68, opened 2025-11-20
Example on how the worker works with coordinator

BinaryFiddler:
Run all services:
```
docker compose up
```

Produce sync action data and grpc into the coordinator
```
docker compose --profile coordinator_test_data up coordinator-test-data-producer -d
```

Worker is successfully reading from the bi-directional stream:

https://github.com/user-attachments/assets/77c1b8ee-3160-4741-bfa0-17fbd8519a2f

Druid ingestion is functional, UI works properly:

Screenshot 2025-11-19 at 5 08 33 PM

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -102,14 +102,22 @@ services:
         condition: service_healthy
       minio-bucket-init:
         condition: service_completed_successfully
+      etcd:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+      osprey_coordinator:
+        condition: service_started
     ports:
       - "5001:5000"
     command: ["osprey-worker"]
     environment:
       - PYTHONPATH=/osprey
       - PORT=5000
+      - ETCD_PEERS=http://etcd:2379
       - POSTGRES_HOSTS={"osprey_db":"postgresql://osprey:FoolishPassword@postgres:5432/osprey"}
-      - OSPREY_INPUT_STREAM_SOURCE=kafka
+      # - OSPREY_INPUT_STREAM_SOURCE=kafka
+      - OSPREY_INPUT_STREAM_SOURCE=osprey_coordinator
       - OSPREY_STDOUT_OUTPUT_SINK=True
       - OSPREY_KAFKA_BOOTSTRAP_SERVERS=["kafka:29092"]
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -251,6 +259,30 @@ services:
       - /bin/bash
     command: ["/osprey/example_data/generate_test_data.sh"]
 
+  # Optional coordinator test data generator - run with:
+  # docker compose --profile coordinator_test_data up coordinator-test-data-producer -d
+  coordinator-test-data-producer:
```
BinaryFiddler, line 264: Largely a clone of how to test with kafka input stream.

docker-compose.yaml
```diff
@@ -102,14 +102,22 @@ services:
...
       - "5001:5000"
     command: ["osprey-worker"]
     environment:
       - PYTHONPATH=/osprey
       - PORT=5000
+      - ETCD_PEERS=http://etcd:2379
       - POSTGRES_HOSTS={"osprey_db":"postgresql://osprey:FoolishPassword@postgres:5432/osprey"}
-      - OSPREY_INPUT_STREAM_SOURCE=kafka
+      # - OSPREY_INPUT_STREAM_SOURCE=kafka
+      - OSPREY_INPUT_STREAM_SOURCE=osprey_coordinator
```
BinaryFiddler, line 120: Probably shouldn't leave this as default. I'm open to suggestions on how we make this more visible and less confusing

example_data/generate_coordinator_test_data.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+# continuously generate and send test actions to the osprey coordinator via gRPC
+# this mimics the Kafka test data generator but sends directly to the coordinator
+
+set -e
+
+COORDINATOR_HOST="${COORDINATOR_HOST:-localhost:19951}"
```
BinaryFiddler, line 8: This is almost the same as the `generate_test.sh`, it uses the same template. Maybe the scripts can be merged.

docker-compose.yaml
juanmrad: can't you make this an env variable? with a default? `${OSPREY_INPUT_STREAM_SOURCE:-kafka}`

BinaryFiddler:
https://github.com/roostorg/osprey/pull/70

PR #70, opened 2025-11-23
Add dockerignore, add kafka support for coordinator

BinaryFiddler:
- Coordinator now supports Kafka as input
- Remove coordinator from the base docker compose
- Add some documentation on how to run osprey with coordinator
- Add a .dockerignore file to speed up local build


osprey_coordinator/src/kafka.rs
```diff
@@ -0,0 +1,322 @@
...
+use serde::Deserialize;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::time::{timeout, Duration as TokioDuration, Instant};
+
+use crate::proto::Action as OspreyProtoAction;
+
+async fn decode_proto_message(
```
BinaryFiddler: This is the same as the one for `pubsub`, maybe pull it out to as a utils fn?

osprey_coordinator/src/kafka.rs
```diff
@@ -0,0 +1,305 @@
+use crate::{
```
BinaryFiddler: referring the pubsub implementation: https://github.com/roostorg/osprey/blob/binaryfiddler/coordinator-kafka/osprey_coordinator/src/pubsub.rs

osprey_coordinator/src/main.rs
```diff
@@ -96,7 +97,12 @@ async fn main() -> Result<()> {
         ));
 
     let pubsub_fut = start_pubsub_subscriber(
-        snowflake_client,
+        snowflake_client.clone(),
+        priority_queue_sender.clone(),
+        metrics.clone(),
+    );
+    let kafka_fut = kafka::start_kafka_consumer(
+        snowflake_client.clone(),
```
cmttt: I thought the implementation we agreed upon was:

Let the user choose kafka or pubsub (probably through an env variable), and then only initialize the subscriber/consumer for the one they chose.

osprey_coordinator/src/kafka.rs
cmttt: Yeah, we should be able to decouple common utils from the underlying queue utils. In my mind, the "queue" is just a plugin for reading messages. Once the message is received, we should handle it the same regardless of where we are at.

osprey_coordinator/src/kafka.rs
```diff
@@ -0,0 +1,302 @@
...
```
cmttt: There's a lot of duplication between the Kafka consumer and the pubsub consumer. My suggestion is that we break this out in to a trait and implementing a consumer struct for each Kafka/Pubsub. 

This is claude generated framework for what this would look like:

```
// A trait for any message consumer
#[async_trait]
pub trait MessageConsumer {
    type Message: ConsumerMessage;
    type Error: std::error::Error + Send + Sync + 'static;
    
    /// Receive the next message
    async fn receive(&mut self) -> Result<Self::Message, Self::Error>;
    
    /// Acknowledge a message
    async fn ack(&self, message: &Self::Message) -> Result<(), Self::Error>;
    
    /// Negative acknowledge a message (for redelivery)
    async fn nack(&self, message: &Self::Message) -> Result<(), Self::Error>;
}

// A trait for messages from any consumer
pub trait ConsumerMessage {
    fn data(&self) -> &[u8];
    fn attributes(&self) -> &HashMap<String, String>;
    fn timestamp(&self) -> Timestamp;
    fn id(&self) -> String;
}

// Then implement these for Kafka and PubSub
pub struct KafkaConsumer { /* ... */ }
pub struct PubSubConsumer { /* ... */ }

impl MessageConsumer for KafkaConsumer { /* ... */ }
impl MessageConsumer for PubSubConsumer { /* ... */ }

pub async fn start_consumer<C: MessageConsumer>(
    consumer: C,
    snowflake_client: Arc<SnowflakeClient>,
    priority_queue_sender: PriorityQueueSender,
    metrics: Arc<OspreyCoordinatorMetrics>,
    config: ConsumerConfig,
) -> Result<()> {
    // Common logic for receiving, decoding, sending to queue, ack/nack
}
```

You might need to change a few things as you go along, but the general skeleton is there. If we add another consumer down the line, they would just need to implement `MessageConsumer` instead of putting all the logic in a bespoke way like this.

cmttt: Also `consumers` could go in a separate folder under `src`.

BinaryFiddler: Yes, this is a good callout. There is definitely room for better abstractions here. I'll work on improving it.

osprey_coordinator/src/main.rs
BinaryFiddler: Could there be a case where they might use both at the same time?

BinaryFiddler: synced offline, only support one or the other

osprey_coordinator/src/kafka.rs
BinaryFiddler: I left this part out: 

```
pub async fn start_consumer<C: MessageConsumer>(
    consumer: C,
    snowflake_client: Arc<SnowflakeClient>,
    priority_queue_sender: PriorityQueueSender,
    metrics: Arc<OspreyCoordinatorMetrics>,
    config: ConsumerConfig,
) -> Result<()> {
    // Common logic for receiving, decoding, sending to queue, ack/nack
}
```

as pubsub and kafka seem to have pretty different internal looping implementations that I'd leave alone from this PR if that's okay.

docker-compose.yaml
```diff
@@ -258,7 +258,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
-      - metadata_data:/var/lib/postgresql
```
EXBreder, line 261: intentional?

osprey_coordinator/src/main.rs
```diff
@@ -95,11 +95,43 @@ async fn main() -> Result<()> {
...
+            Box::pin(start_pubsub_subscriber(
+                snowflake_client.clone(),
+                priority_queue_sender.clone(),
+                metrics.clone(),
+            ))
+                as std::pin::Pin<Box<dyn std::future::Future<Output = Result<()>> + Send>>
+        }
+        _ => {
+            tracing::warn!(
+                "invalid OSPREY_COORDINATOR_CONSUMER_TYPE '{}', defaulting to Kafka",
```
EXBreder: if invalid should just hard fail. If empty, its probably fine to default.

example_docker_compose/run_osprey_with_coordinator/docker-compose.yaml
```diff
@@ -0,0 +1,463 @@
...
+      kafka:
+        condition: service_healthy
+    command: >
+      bash -c "
+        kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic osprey.actions_input --partitions 3 --replication-factor 1 &&
+        kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic osprey.execution_results --partitions 3 --replication-factor 1 &&
+        kafka-topics --bootstrap-server kafka:29092 --list
+      "
+
+  osprey-worker:
```
EXBreder: Heavily recommend we don't just copy and past the entire yaml from the main one, its really easy to forget to modify one thing in here if the main one also gets modified.

Instead:
1. consider this one to be an "override" -> `docker-compose.coordinator.yaml`, and just have the new/changed things listed.
2. change the readme to include these files in the docker compose up command -> `docker compose up -f docker-compose.yaml -f location/to/docker-compose.coordinator.yaml up`
3. add a helper script in the root to swap between them (such as `start.sh --with-coordinator`)

osprey_coordinator/src/consumer/pubsub.rs
EXBreder: this is a good opportunity to fix these tests

docker-compose.yaml
BinaryFiddler: yup, this is already in main. `postgres 18+` change where the db volume is mounted

osprey_coordinator/test_data/pubsub_proto_message.json
```diff
@@ -0,0 +1,12 @@
+[
```
BinaryFiddler, line 1: I took the structure from the discord repo and used a script to generate this file:

```
import base64
from osprey.rpc.actions.v1 import action_pb2, action_types_pb2

# Create an Action message
action = action_pb2.Action()
action.id = 987654321
action.guild_created.CopyFrom(action_types_pb2.GuildCreated())  # Empty GuildCreated message

# Encode to bytes and base64
proto_bytes = action.SerializeToString()
base64_encoded = base64.b64encode(proto_bytes).decode('utf-8')
print(base64_encoded)
```

EXBreder (approved):

PR #69, opened 2025-11-20
Run precommit in CI as documented in README

ryanprior:
## Summary

README.md and DEVELOPMENT.md both indicate that devs should run `uv run pre-commit run --all-files` as part of the standard dev workflow. Regrettably, this fails on the main branch.

At present, CI runs an abridged version of the checks documented in the dev workflow, leading to a lack of friction for PRs that fail pre-commit checks.

## In this PR
- updates files changed by pre-commit
- adds pre-commit into the Python code quality job in CI
  + avoids running duplicate tasks by skipping what was already run & using the GitHub Actions cache

## Next steps
I designed this PR to be a minimal change to the current CI behavior, but you might consider these follow-ons steps to save some CI time:
- run ruff & mypy through pre-commit to take advantage of caching
- use `--from-ref FROM_REF` + `--to-ref TO_REF` to limit code quality checks to changed files only

Saving CI time here may be low priority, since the integration tests are by far the slowest to run, leaving the Python code quality tests well outside the critical path.

PR #72, opened 2025-11-24
fix: ci code quality to use pre-commit

caidanw:
Summary
---
Thanks to the PR #69 for pointing this out.

I've removed two rules that were causing more trouble than they were solving (specifically around trailing lines in files).

Additionally I've migrated the python code quality job to use pre-commit instead of running each tool on it's own.

Also formatted some JSON files that were somehow not previously formatted.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -32,14 +32,19 @@ jobs:
       - name: Install dependencies
         run: uv sync --dev
 
-      - name: Run ruff linting
-        run: uv run ruff check .
+      - name: Set PY
+        run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+      - uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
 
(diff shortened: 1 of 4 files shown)


.github/workflows/code-quality.yml
```diff
@@ -32,14 +32,19 @@ jobs:
       - name: Install dependencies
         run: uv sync --dev
 
-      - name: Run ruff linting
-        run: uv run ruff check .
-
-      - name: Run ruff formatting check
-        run: uv run ruff format --check .
+      - name: Set PY
```
BinaryFiddler, line 35: What does `Set PY` mean? Setting the python version?

BinaryFiddler (approved):

.github/workflows/code-quality.yml
ryanprior: I took this from the documentation for precommit. It takes the python version output and hashes it, giving you an arbitrary stable identifier for a particular build of Python. Their example called it PY so that's what I went with, but the name has no further significance.

ryanprior (approved):
This looks solid. I did not have the context when writing #69 to know whether you'd want to keep those problematic lint steps or not, but removing them is tidy.

.github/workflows/code-quality.yml
caidanw: Yep, @ryanprior is correct

PR #74, opened 2025-11-25
pin postgres version to 18

BinaryFiddler:
Pinning postgres image version to 18.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -254,11 +254,11 @@ services:
   postgres:
     hostname: postgres
     container_name: postgres
-    image: postgres:latest
+    image: postgres:18
     ports:
       - "5432:5432"
     volumes:
-      - metadata_data:/var/lib/postgresql/data
+      - metadata_data:/var/lib/postgresql
     environment:
       - POSTGRES_PASSWORD=FoolishPassword
       - POSTGRES_USER=osprey
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -258,7 +258,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
-      - metadata_data:/var/lib/postgresql/18/docker
+      - metadata_data:/var/lib/postgresql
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -254,11 +254,11 @@ services:
   postgres:
     hostname: postgres
     container_name: postgres
-    image: postgres:latest
+    image: postgres:18
     ports:
       - "5432:5432"
     volumes:
-      - metadata_data:/var/lib/postgresql/data
+      - metadata_data:/var/lib/postgresql
```
ThisIsMissEm: This also isn't the correct path for postgresql 18: https://github.com/docker-library/docs/blob/master/postgres/README.md#pgdata

```suggestion
      - metadata_data:/var/lib/postgresql/18/docker
```

haileyok (approved):
lgtm after @ThisIsMissEm's note

docker-compose.yaml
BinaryFiddler: I think it should be `/var/lib/postgresql` @ThisIsMissEm 

`/var/lib/postgresql/18/docker` looks like the internal data directory path
`/var/lib/postgresql` is the mount point where the volume is attached from the host

postgres will automatically create and use /var/lib/postgresql/18/docker inside that mounted volume for its data.

BinaryFiddler: The doc also suggests this:

> Important Change: https://github.com/docker-library/postgres/pull/1259. For 18 it is /var/lib/postgresql/18/docker. Later versions will replace 18 with their respective major version (e.g., /var/lib/postgresql/19/docker for PostgreSQL 19.x). **The defined VOLUME was changed in 18 and above to /var/lib/postgresql. Mounts and volumes should be targeted at the updated location.**

specifically: **The defined VOLUME was changed in 18 and above to /var/lib/postgresql. Mounts and volumes should be targeted at the updated location.**

PR #75, opened 2025-11-25
update submit-specs.sh to use correct host

haileyok:
@BinaryFiddler 

Updates `submit-specs.sh` to use the correct Druid coordinator hostname after the docker networking got reworked.


BinaryFiddler (approved):
I wonder how we didn't caught this for a month

PR #83, opened 2025-12-05
build and publish coordinator image

BinaryFiddler:
Test image is built and pushed here:
- https://github.com/roostorg/osprey/pkgs/container/osprey%2Fosprey-coordinator

### DIFF
--- a/.github/workflows/publish-coordinator-image.yml
+++ b/.github/workflows/publish-coordinator-image.yml
@@ -1,0 +1,86 @@
+name: Build and Publish Osprey Coordinator Docker Image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "osprey_coordinator/**"
+      - "proto/**"
+      - ".github/workflows/publish-coordinator-image.yml"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "osprey_coordinator/**"
+      - "proto/**"
+      - ".github/workflows/publish-coordinator-image.yml"
+  release:
+    types: [published]
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}/osprey-coordinator
+
+jobs:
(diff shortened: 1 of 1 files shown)


cmttt (approved):

.github/workflows/publish-coordinator-image.yml
```diff
@@ -0,0 +1,86 @@
...
+      - "osprey_coordinator/**"
+      - "proto/**"
+      - ".github/workflows/publish-coordinator-image.yml"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "osprey_coordinator/**"
+      - "proto/**"
+      - ".github/workflows/publish-coordinator-image.yml"
```
cassidyjames, line 17: I might be misunderstanding, but:
- will this build and publish the image on every PR opened against main?
- if so, …is that right?

The "on push to main" and "on release" make sense to me, but this (on pull_request) was a bit of a red flag. Again, I might just be misunderstanding the workflow, though.

EXBreder: there should be no issues publishing images on each commit/PR to main - its pretty common practice to always publish, as long as the semver is reasonable.

BinaryFiddler: You are right, it runs the workflow for both pull request and on main commit, but it only publishes the image when the commit is made into main: 

https://github.com/roostorg/osprey/pull/83/files#diff-7bc9de45ddfed48a46ead8d69e032d22941792fd547ad0ecf7fb6358f32802d5R41

https://github.com/roostorg/osprey/pull/83/files#diff-7bc9de45ddfed48a46ead8d69e032d22941792fd547ad0ecf7fb6358f32802d5R67

https://github.com/roostorg/osprey/pull/83/files#diff-7bc9de45ddfed48a46ead8d69e032d22941792fd547ad0ecf7fb6358f32802d5R75

EXBreder (approved):

PR #89, opened 2025-12-10
Remove BigTable thread pool wrapper that was causing hangs

cmttt:
The thread pool wrapper around BigTable operations was causing hangs with gevent monkey patching. Since grpc.experimental.gevent.init_gevent() is already called, gRPC is compatible with gevent and can be called directly.

Changes:
- Remove _bigtable_threadpool and ThreadPool import
- Call read_row and mutate_rows directly instead of wrapping in thread pool
- Reduce StoredExecutionResultOutputSink timeout from 10s to 5s now that operations are fast (~50ms) without the thread pool overhead


ayubun (approved):
is this experimental init new ? :D

PR #94, opened 2025-12-12
Fix remaining broken links

bmuenzenmeyer:
Building off of #93, I ran https://www.npmjs.com/package/markdown-link-check now that I am home, the tool internalized by the action I suggested. It found these broken links. 

Two are the result of the move to a github org community files - a good idea! These also could link to the README tab, but I think that's a bit janky and not as clear to the user.


julietshen (approved):
Thank you for your PR!

images/ROOST-Horizontal-Yellow.png
cassidyjames, line 1: Strange that this file had a space at the end! 😅

PR #98, opened 2025-12-15
fix postgres metadata volume

haileyok:
cc @BinaryFiddler, i think this accidentally got reverted back in a subsequent commit

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -258,7 +258,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
Yep, looks like there was some confusion here since #70 predated #74:

https://github.com/roostorg/osprey/pull/70/changes#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9L261-R261

PR #111, opened 2026-01-13
Package osprey rpc as zip file

cmttt:
### Summary

This PR adds infrastructure to automatically build and publish the `osprey-rpc` Python package as release assets whenever a new GitHub release is created. It also regenerates the proto files to fix import issues.

#### New: GitHub Actions Release Workflow
- Added `.github/workflows/release-osprey-rpc.yml` that triggers on release publish
- Builds the `osprey_rpc` sdist using `uv build`
- Creates both `.tar.gz` and `.zip` archives (zip required for downstream pycross/Bazel compatibility)
- Uploads both artifacts to the GitHub release

#### Proto Regeneration
- Regenerated all proto-generated Python files in `osprey_rpc/src/`
- Fixed stale imports that referenced non-existent `osprey.rpc.labels` module
- Added missing `google.*` proto dependencies (protobuf, api, iam, pubsub, bigtable, etc.)

### Why

The `discord_api` service needs to consume `osprey-rpc` as a dependency to communicate with the Osprey coordinator. Discord's build system (Bazel + pycross) requires dependencies to be:
1. Available as archive URLs (not git+ URLs)
2. Distributed as `.zip` files (not `.tar.gz`)

This workflow enables `discord_api` to depend on `osprey-rpc` via:
```toml
[tool.uv.sources]
osprey-rpc = { url = "https://github.com/roostorg/osprey/releases/download/<tag>/osprey_rpc-0.1.0.zip" }
```

### Testing

- [x] Workflow runs successfully on release publish
- [x] Generated `.zip` archive is consumable by downstream projects
- [x] Proto imports are correct (no `ModuleNotFoundError`)
- [x] `uv.lock` regenerated with uv 0.9.24 to match CI

---


EXBreder (approved):

gen-protos.sh
```diff
@@ -11,6 +11,6 @@ shopt -s globstar 2>/dev/null || true
 # change to directory of script
 cd "$(dirname "$0")"
 
-glob=(./proto/**/*.proto)
+glob=(./proto/osprey/**/*.proto)
```
cmttt, line 14: @EXBreder I think this is fine so we don't have to generate all the google protos

PR #117, opened 2026-01-20
Docs: Restructure, simplify, and set up mdbook

cassidyjames:
- Set up mdbook like we have in other repos; once merged, we can deploy a docs site
- Deduplicated info between README.md and DEVELOPMENT.md
- Broke DEVELOPMENT.md down into more digestible sub-pages
- Removed some apparently AI-generated fluff that was not particularly helpful
- Removed references to "not in v0"/"coming in v1"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<img width="200" height="64" alt="ROOST logo" src="images/ROOST-Horizontal-Yellow.png" />
+
 # Osprey
 
 **Automate the obvious and investigate the ambiguous.** High-performance safety rules engine for real-time event processing at scale. 


docs/development/tools.md
cassidyjames, line 1: I'm not actually sure how useful this page is; it feels like what I'd get if I asked Claude to explain the development tools. Do people find it helpful?

julietshen: it was helpful for me as someone who didn't know what any of them are, but i feel like it might be unnecessary? cc @ThisIsMissEm  @haileyok  what do you think as non-OG osprey developers?

docs/development/troubleshooting.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Solution**: Protobuf generated files are excluded in configuration. If you see errors, check that files match the exclusion patterns in `pyproject.toml`.
+
+### Import errors during type checking
+
+**Solution**: Ensure all dependencies are installed:
+
+```bash
+uv sync
+```
```
julietshen, line 32: @cassidyjames do you want to add any of your troubleshooting here too? IE if you run into x problem, it might be because you're using fedora

docs/rules.md
```diff
@@ -1,12 +1,8 @@
-# Osprey Docs
-
-# Osprey Docs
+# Osprey Rules
```
julietshen, line 1: let's get this merged in first, but i would love for some of @haileyok's rule documentation to be added here! it was super helpful.

README.md
```diff
@@ -1,126 +1,47 @@
-# ROOST - Osprey
+# Osprey
 
-<img width="200" height="64" alt="Copy of ROOST-Mark-Yellow" src="/images/ROOST-Horizontal-Yellow.png" />
```
julietshen, line 3: nit: can we keep the ROOST horizontal image

docs/development/troubleshooting.md
cassidyjames: Weirdly when trying to reproduce on another machine, I didn't get the same issues! So I'm not sure if it was just specific to something on that machine. I opened the issue in the repo where it can be further investigated, and it would make sense to add any further gotchas here, but I think at the moment this is good.

docs/development/tools.md
ThisIsMissEm: I don't think it hurts to have. Being explicit about the tools you use for development and what they're used for is often a good thing.

README.md
cassidyjames: Sure, not a big deal. I think ideally we could have like a **ROOST | Osprey** lockup or something but we can leave the ROOST logo for now.

docs/rules.md
haileyok: yea once we get this in i'll make some additional PRs for that documentation!

docs/development/troubleshooting.md
```diff
@@ -0,0 +1,32 @@
...
+### MyPy errors on protobuf files
+
+**Solution**: Protobuf generated files are excluded in configuration. If you see errors, check that files match the exclusion patterns in `pyproject.toml`.
+
+### Import errors during type checking
+
+**Solution**: Ensure all dependencies are installed:
+
+```bash
+uv sync
```
haileyok, line 31: i think this should be `uv sync --frozen`

PR #120, opened 2026-01-20
Single script to run with curl command to run the osprey demo

vinaysrao1:
Add a demo script that runs with a single curl command to install osprey, start services and the example event generator. It will check for existing docker services and port collisions. This overlaps a bit with start.sh. Command:

curl -sSL https://raw.githubusercontent.com/roostorg/osprey/main/demo.sh | bash

### DIFF
--- a/demo.sh
+++ b/demo.sh
@@ -66,6 +66,32 @@ check_port_available() {
     return 0
 }
 
+check_existing_services() {
+    echo -e "${YELLOW}Checking for existing Osprey services...${NC}"
+
+    # Check for running containers from this compose project
+    local running_containers=$(docker ps --filter "name=osprey\|kafka\|postgres\|minio\|druid\|zookeeper\|snowflake" --format "{{.Names}}" 2>/dev/null | head -10)
+
+    if [ -n "$running_containers" ]; then
+        echo -e "${YELLOW}Found existing services that may conflict:${NC}"
+        echo "$running_containers" | while read -r container; do
+            echo -e "  • $container"
+        done
+        echo ""
+        echo -e "${YELLOW}Starting the demo will stop these services and remove their volumes (data will be lost).${NC}"
+        echo -n -e "${BLUE}Do you want to continue? [y/N]: ${NC}"
+        read -r response
+        if [[ ! "$response" =~ ^[Yy]$ ]]; then
+            echo -e "${RED}Aborted.${NC}"
+            exit 0
+        fi
+        echo ""
+    else
+        echo -e "${GREEN}✓ No conflicting services found${NC}"
+    fi
+    echo ""
+}
+
 check_required_ports() {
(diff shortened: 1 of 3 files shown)


cassidyjames:
A note for anyone testing, the curl-pipe-bash flow for the latest version of the script from this branch can be used with:

```sh
curl -sSL https://raw.githubusercontent.com/roostorg/osprey/osprey-add-single-start-script/demo.sh | bash
```

It will be available at `https://raw.githubusercontent.com/roostorg/osprey/main/demo.sh` once merged (assuming it's merged).

demo.sh
```diff
@@ -0,0 +1,290 @@
...
+DOCKER_CLI_HINTS=false docker compose -f "$COMPOSE_FILE" --progress=quiet up -d --build --quiet-pull
+echo -e "${GREEN}✓ Core services starting${NC}"
+echo ""
+
+# Step 4: Wait for critical services
+echo -e "${YELLOW}Step 4: Waiting for services to be ready...${NC}"
+echo ""
+
+# Wait for infrastructure containers (use Docker health checks)
+wait_for_container "kafka" 90 || exit 1
```
haileyok: should merge in main and update this, since this got renamed to `osprey-kafka` now
https://github.com/roostorg/osprey/commit/5210214d7a1ced8cff76b6332174f8067cecb49c

demo.sh
```diff
@@ -0,0 +1,290 @@
...
+wait_for_http_service "Osprey UI API" "http://localhost:5004/config" 120 || exit 1
+wait_for_http_service "Osprey UI" "http://localhost:5002" 90 || exit 1
+
+echo ""
+echo -e "${GREEN}✓ All services are ready${NC}"
+echo ""
+
+# Step 5: Start test data generator
+echo -e "${YELLOW}Step 5: Starting test data generator...${NC}"
+DOCKER_CLI_HINTS=false docker compose -f "$COMPOSE_FILE" --progress=quiet --profile test_data up -d kafka-test-data-producer
```
haileyok: oh also this one, it has a `osprey-` prefix now too

haileyok:
> Testing on Linux (Fedora Silverblue here), I get a failure on step 4:

seems like this was the out of date kafka container names. i also got this when i first ran via curl, but worked if i updated them. tested on ubuntu

haileyok (approved):
works for me after updating the two lines pointed out above. 

i also might suggest trapping `INT` and `TERM` signals and having some cleanup happen (`docker compose down`) if someone exits early or something. if anything goes wrong you'll end up with all the services still running (which in turn prevents future executions of the script). probably for any early exits that occur after the git clone step?

haileyok (approved):
yep this is looking good. good call on removing bigtable, i think it was one of the more annoying things/things that took a while to actually start up/stop

PR #121, opened 2026-01-21
add check censored udf, tokenize udf

haileyok:
I originally wrote this without a translation table, but decided that because the produced regex were so large that it would be best to just create that table instead. The functionality of this PR and the previous one is the same, though
with fairly reduced complexity. It adds two new string UDFs:

- StringTokenize, which converts the given text into a list of individual tokens (split at whitespace or punctuation). Not strictly necessary for this PR, but useful for it and feels like a good time to add it in
- CheckCensored, which builds a regex for a given input phrase then checks a given input token against the regex

There's some existing functionality for lookalikes in `string.py`, specifically the `StringClean` UDF. There's a variety of additional things that I've added in this new UDF though (that have also come up quite a bit in the wild)

- Still will match terms that attempt to obfuscate with "separator" characters, i.e. using "c___a___t" to try and get around matching for "cat"
- Handles zero-width spaces that are often used for the same purpose
- Handles a larger list of characters
- Allows for matching _only_ when a given term is obfuscated, i.e. it's fine to use "cat" but not okay to use "<4t"

There's likely still a bit of overhead from using regex here to achieve some more useful results from the current implementation of this, but we get some additional info/control by using them.

### DIFF
--- a/osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_strings.py
+++ b/osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_strings.py
@@ -337,7 +337,8 @@ def test_extract_urls(execute: ExecuteFunction, text: str, expected_result: List
         ('hello, world!', ['hello', 'world']),
         ('end. start', ['end', 'start']),
         ('café résumé', ['café', 'résumé']),
-        ("don't", ["don't"]),  # curly apostrophe
+        ('donʼt', ["don't"]),  # curly apostrophe (u02bc)
+        ('don’t', ["don't"]),  # curly apostrophe (u2019)
         ("cat's", ["cat's"]),
         ("''hello", ['hello']),
         ("test''test", ['test', 'test']),


osprey_worker/src/osprey/engine/stdlib/udfs/string.py
```diff
@@ -400,3 +400,28 @@ def execute(self, execution_context: ExecutionContext, arguments: StringArgument
 
         # return any valid urls encountered in the message
         return list(valid_urls)
+
+
+_TOKEN_PATTERN = re.compile(r"[\w]+(?:'[\w]+)?", re.UNICODE)
+
+
+def tokenize_text(s: str) -> list[str]:
+    s = s.replace("'", "'").replace('ʼ', "'")
```
BinaryFiddler: Actually, what does first `replace` do?

haileyok: ah nice catch. there are two different apostrophes we want to replace with a "normal" apostrophe, and it looks like that first one actually might have gotten replaced with a normal apostrophe maybe through copy/paste or something. switched it to just use unicode escapes so its more obvious what it's doing

https://github.com/roostorg/osprey/pull/121/changes/a5fab61f8034770da163cc85716bf9db7af55c24

haileyok: + test https://github.com/roostorg/osprey/pull/121/changes/103d03fc9d2023769dffe8cdaf19d49447d00f02

BinaryFiddler: nice!

PR #125, opened 2026-01-23
Add safeparse to string udfs

cmttt:
fixes the string udfs


ayubun (approved):

EXBreder (approved):
Note that this adds ipv6 support for domains too, which isn't really mentioned in the PR.

Looks fine, despite me having concerns with the fact that URLs _can_ be w/o a protocl and be valid.

PR #128, opened 2026-01-29
Add error_on_empty parameter to HasLabel UDF

cmttt:
## Summary
- Add optional `error_on_empty` parameter (default `False`) to `HasLabel` UDF
- When `True`, raises `EmptyEntityError` if the entity has no labels at all
- Allows fail-closed behavior for safety-critical label checks

## Motivation
When the labels service is degraded and returns empty default responses instead of actual label data, rules that check `not HasLabel(...)` could incorrectly fire. This parameter provides an opt-in mechanism to detect and fail on potential data fetch failures.

## Usage
```python
# Fail-closed: raise EmptyEntityError if entity has zero labels
HasLabel(entity=SomeEntity, label="some_label", error_on_empty=True)
```

## Important: When to use this parameter

**Only use `error_on_empty=True` when ALL of these conditions apply:**

1. A false negative from `HasLabel` could cause **dangerous rule evaluations** (e.g., incorrectly allowing a known-bad entity through safety checks)
2. You are checking `not HasLabel(...)` where empty data would cause the rule to incorrectly fire

**Do NOT use this for:**
- General label checks where empty is a valid state
- Non-safety-critical rules

## Why not return None instead?
`HasLabel` returns `bool`. Returning `None` would change the signature to `Optional[bool]`, requiring all callers to handle the `None` case and risking subtle bugs where `None` gets coerced to `False`. The exception approach explicitly fails the rule execution rather than silently propagating an unknown state.

## Test plan
- [x] Added unit tests for error_on_empty behavior
- [x] All existing label tests pass

### DIFF
--- a/osprey_worker/src/osprey/engine/stdlib/udfs/labels.py
+++ b/osprey_worker/src/osprey/engine/stdlib/udfs/labels.py
@@ -107,6 +107,19 @@ class _SimpleStatus(Enum):
     REMOVED = 'removed'
 
 
+class EmptyEntityError(Exception):
+    """Raised when an entity has no labels and error_on_empty is True."""
+
+    def __init__(self, entity: EntityT[Any], label: str):
+        self.entity = entity
+        self.label = label
+        super().__init__(
+            f"Entity '{entity.type}/{entity.id}' has no labels. "
+            f'This may indicate the labels service failed to fetch data. '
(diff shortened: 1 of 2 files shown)


ayubun (approved):
lgtm, thank u leon!

EXBreder (approved):

PR #129, opened 2026-01-30
Fix MXLookup UDF timeout errors

cmttt:
## Summary

- Configure DNS resolver with 0.5s query timeout and 1.0s lifetime (was using defaults which could take 30+ seconds)
- Add 2.0s gevent timeout wrapping entire execute method (covers both MX + A lookups)
- Add metrics: `mx_lookup.execute.{success,dns_timeout,gevent_timeout,no_record}`
- Catch `DNSTimeout` explicitly and convert to `ExpectedUdfException`

## Test plan

- [ ] Added test for DNS timeout scenario
- [ ] Deploy to staging and monitor `mx_lookup.execute.*` metrics
- [ ] Verify timeout errors decrease significantly

### DIFF
--- a/osprey_worker/src/osprey/engine/stdlib/udfs/mx_lookup.py
+++ b/osprey_worker/src/osprey/engine/stdlib/udfs/mx_lookup.py
@@ -1,11 +1,23 @@
+import gevent
+from dns.exception import Timeout as DNSTimeout
 from dns.resolver import NXDOMAIN, YXDOMAIN, LRUCache, NoAnswer, NoNameservers, Resolver
 from osprey.engine.executor.execution_context import ExpectedUdfException
+from osprey.worker.lib.instruments import metrics
 
 from ._prelude import ArgumentsBase, ExecutionContext, UDFBase
 from .categories import UdfCategories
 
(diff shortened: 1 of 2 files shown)


osprey_worker/src/osprey/engine/stdlib/udfs/mx_lookup.py
```diff
@@ -21,11 +33,20 @@ class MXLookup(UDFBase[Arguments, str]):
 
     def execute(self, execution_context: ExecutionContext, arguments: Arguments) -> str:
         try:
-            mx_answer = resolver.resolve(arguments.domain, 'MX', raise_on_no_answer=True)[0]
-            domain = mx_answer.exchange.to_text()
-            a_record_answers = resolver.resolve(domain, 'A', raise_on_no_answer=True)
-        except (NoAnswer, NXDOMAIN, YXDOMAIN, NoNameservers):
+            with gevent.Timeout(MXLOOKUP_GEVENT_TIMEOUT, False):
```
haileyok, line 36: i think you don't want `False` there if you want to catch the exception below

haileyok (approved):

PR #139, opened 2026-02-13
Defer REGISTRY import to fix empty validator snapshot

cmttt:
## Summary
Move the `REGISTRY` import from module-level to inside `parse_query_to_validated_ast()`.

## Problem
The query-language `REGISTRY` in `ast_validator.py` snapshots `ValidatorRegistry.get_instance()` at module import time. This import is triggered during engine bootstrap (via `plugin_manager` import chain) before `bootstrap_ast_validators()` populates the singleton. The snapshot captures an empty registry, so `DruidQueryTransformer.transform_Call` fails with `KeyError` on queries with UDF calls (e.g. `DidAddLabel(...)`).

## Fix
Defer the `REGISTRY` import in `__init__.py` from module-level to inside `parse_query_to_validated_ast()`. The snapshot is then taken at first call (request time) when validators are already registered. One line removed, one line added.

## Test plan
- [ ] Existing query_language tests pass
- [ ] Existing ast_druid_translator tests pass


EXBreder (approved):

jani745765-netizen (approved):

cassidyjames:
@EXBreder I noticed this has been sitting here approved; is there any reason not to merge it?

PR #141, opened 2026-02-19
fix build problems with updated uv install on macos

### LINKED ISSUE #136: Docker build fails on fresh build - needs grpcio-tools upgrade
I had cleared my Docker out to test fresh builds and I was using Claude Code to get Osprey up and running to generate a screenshot, but ran into issues. It might be a breaking change introduced by recent uv versions that affects anyone trying to build Osprey from docker right now.

**Problem Description**
Docker build fails with recent uv versions due to grpcio-tools 1.49.1 missing pkg_resources build dependency

When building the Osprey Docker containers, the build fails during dependency installation with the following error:

  × Failed to build `grpcio-tools==1.49.1`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

ModuleNotFoundError: No module named 'pkg_resources'

 ### Root Cause:
  - grpcio-tools==1.49.1 has an old setup.py that imports pkg_resources without declaring it as a build dependency
  - Recent versions of uv (0.9.0+ and 0.10.x) enforce stricter build isolation
  - The error occurs regardless of whether setuptools is installed globally
  - When I did a fresh build from scratch, I automatically upgraded to uv 0.10

**Expected behavior**
Osprey build is successful after running docker-compose. 

**Repro steps**
  1. Fresh build of the repo
  2. Run ./start.sh or docker compose up -d
  3. Build fails on grpcio-tools

**Screenshots/Recordings/Logs**

**Context/Environment (please complete the following information):**
 - ARM Mac

**Additional context**
- CI: version: "latest" (always gets the newest uv, in this case the newer ones that introduced a breaking change)
  - Dockerfile: pip install --upgrade pip uv (always gets the newest uv)
  - Documentation: Just says to install uv, no version specified

### recommendation by Claude
  1. to a newer version that properly declares build dependencies
  2. Or: Modify Dockerfile to use --no-build-isolation flag (less secure but works)
  3. Or: Pin a specific older version of uv that works and regenerate lockfile
  

[osprey_build_error.log](https://github.com/user-attachments/files/25276023/osprey_build_error.log)

haileyok:
fixes #136, though not actually sure if it needs to be this elaborate. im assuming there was a reason for pinning grpcio on x86_64 to 1.49.1 before though, so leaving that here as is. also had to bump status and tools, since the same error happens if you dont.

someone should sanity check this probably

cc @julietshen also worth seeing if this fixes things for you

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,12 +44,14 @@ common = [
     "graphviz==0.20",
     "greenlet",
     "grpc-google-iam-v1==0.14.2",
-    "grpcio==1.49.1; platform_machine == 'x86_64'",  # Normal dependency for production
-    "grpcio==1.53.*; platform_machine == 'aarch64'", # Only install on ARM64 architecture for local development on MacOS
+    "grpcio==1.49.1; platform_machine == 'x86_64'",
+    "grpcio==1.53.*; platform_machine != 'x86_64'",
     "grpcio-health-checking==1.44.0",
     "grpcio-reflection==1.44.0",
-    "grpcio-status==1.49.1",
-    "grpcio-tools==1.49.1",
+    "grpcio-status==1.49.1; platform_machine == 'x86_64'",
+    "grpcio-status==1.53.*; platform_machine != 'x86_64'",
+    "grpcio-tools==1.49.1; platform_machine == 'x86_64'",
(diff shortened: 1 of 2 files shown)


julietshen (approved):
LGTM

PR #145, opened 2026-02-21
update/add to rules documentation

haileyok:
Making some changes for clarity to the rules writing documentation as well as adding some new information like structuring rules and models nicely, writing a little example rule with the full flow, etc.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -135,7 +135,7 @@ ReplyId: Entity[str] = JsonData(
 )
 ```
 
-The `JsonData` UDF (more on UDFs to follow) lets us take the event's JSON and define features based on the contents of that JSON. These features can then be referenced in other rules that we import the `models/record/post.sml` model into. If you have any values inside your JSON object that may not always be present, you can set `required` to `False`, and these features will be `None` whenever the feature is not present.
+The [`JsonData` UDF](#user-defined-functions-udfs) lets us take the event's JSON and define features based on the contents of that JSON. These features can then be referenced in other rules that we import the `models/record/post.sml` model into. If you have any values inside your JSON object that may not always be present, you can set `required` to `False`, and these features will be `None` whenever the feature is not present.
 
 Note that we did not actually create any features for things like `userId` or `handle`. That is because these values will be present in *any* event. It wouldn't be very nice to have to copy these features into each event type's model. Therefore, we will actually create a `base.sml` model that defines these features which are always present. Inside of `models/base.sml`, let's define these.
 
@@ -163,7 +163,7 @@ AccountAgeSeconds: int = JsonData(
 )
 ```
 
-Here, instead of simply using `JsonData`, we instead use the `EntityJson` UDF for the `UserID`. More on this later, but as a rule of thumb, you likely will want to have values for things like a user's ID set to be entities. This will help more later, such as when doing data explorations within the Osprey UI.
+Here, instead of simply using `JsonData`, we instead use the `EntityJson` UDF for the `UserID`. This is covered in the [UDFs section](#user-defined-functions-udfs), but as a rule of thumb, you likely will want to have values for things like a user's ID set to be entities. This will help more later, such as when doing data explorations within the Osprey UI.
 
 ### Model Hierarchy
 
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -239,7 +239,7 @@ def push(self, result: ExecutionResult) -> None:
 
 ## User Defined Functions (UDFs)
 
-User Defined Functions (UDFs) are plugins that enable users of Osprey to extend and customize their use of the Osprey SML. UDFs are implemented as Python functions and are registered
+User Defined Functions (UDFs) are plugins written in Python that enable users of Osprey to extend and customize their use of the Osprey SML. UDFs are implemented as Python functions and are registered
 as a plugin. They extend the `UDFBase` abstract base class with a set of arguments and an output. These will be executed whenever called in SML.
 
 ```python
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -407,3 +407,86 @@ Require(rule=f'actions/{ActionName}.sml')  # will execute 'actions/foo.sml'
 
 Require(rule='ai_services/my_ai_service.sml', require_if=ActionName == "register")
 ```
+
+## Full Example
+
+The following is a complete walkthrough of writing a rule using the project structure described above. The goal is to flag accounts whose first post mentions at least one user and includes a link.
+
+### Writing the Rule
+
+We'll create `rules/record/post/first_post_link.sml` for the rule logic. This file defines both the conditions that cause the rule to evaluate to `True` and the actions to take when it does.
+
+```python
+# First, import the models that you will need inside of this rule
+Import(
+    rules=[
+        'models/base.sml',
+        'models/record/post.sml',
+    ],
+)
+
+# Next, define a variable that uses the `Rule` UDF
+FirstPostLinkRule = Rule(
+    # Set the conditions in which this rule will be `True`
+    when_all=[
+        PostCount == 1, # if this is the user's first post
+        EmbedLink != None, # if there is a link inside of the post
+        ListLength(list=MentionIds) >= 1, # if there is at least one mention in the post
+    ],
+    description='First post for user includes a link embed',
+)
+
+# Finally, set which effect UDFs will be triggered
+WhenRules(
+    rules_any=[FirstPostLinkRule],
+    then=[
+        ReportRecord(
+            entity=PostId,
+            comment='This was the first post by a user and included a link',
+            severity=3,
+        ),
+    ],
+)
+```
+
+### Wiring Up the Rule
+
+We want this rule to run *only* when the event is a post event. Using the project structure described above, this involves three files.
+
+First, `main.sml` at the project root includes a single `Require` statement pointing to the top-level rules index:
+
+```python
+Require(
+    rule='rules/index.sml',
+)
+```
+
+Next, `rules/index.sml` conditionally requires the post rules when the event type matches:
+
+```python
+Import(
+    rules=[
+        'models/base.sml',
+    ],
+)
+
+Require(
+    rule='rules/record/post/index.sml',
+    require_if=EventType == 'userPost',
+)
+```
+
+Finally, `rules/record/post/index.sml` requires the new rule:
+
+```python
+Import(
+    rules=[
+        'models/base.sml',
+        'models/record/post.sml',
+    ],
+)
+
+Require(
+    rule='rules/record/post/first_post_link.sml',
+)
+```
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -206,7 +206,7 @@ to place any effects toward the bottom of workflows.
 
 ## Output Sinks
 
-After all rules are evaluated for an input event, a set of output sinks takes the resulting `ExecutionOutput` and performs additional work based on that data. These may be defined
+After all rules are evaluated for an input event, a set of output sinks takes the resulting `ExecutionResult` and performs additional work based on that data. These may be defined
 as part of a plugin for performing domain specific work.
 
 Some default use cases include a `StdoutOutputSink` which simply outputs the result to the log, a `KafkaOutputSink` which pipes data to Kafka (used for Osprey UI), or the
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -85,7 +85,7 @@ example-rules/
 
 This sort of structure lets you define rules and models that are specific to certain event types so that only the necessary rules are run for various event types. For example, you likely have some rules that should only be run on a `post` event, since only a `post` will have features like `text` or `mention_count`.
 
-Inside of each directory, you may maintain an `index.sml` file that will define the conditional logic in which the rules inside that directory are actually included for execution. Although you could handle all of this conditional logic inside of a single file, maintaining separate `index.sml`s per directory greatly helps with neat organization.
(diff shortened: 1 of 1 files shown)


docs/rules.md
```diff
@@ -4,17 +4,18 @@
...
 
-Rules by themselves only create variables, and without a corresponding `WhenRules()` function call, the rule will have no effects outside of evaluation and query functionality.
+By themselves, rules only create variables, and without a corresponding `WhenRules()` function call, the rule will have no effects outside of evaluation and query functionality.
 
-Rules at present support the following concepts through the `Rule()` function of the same name.
+Rules currently support the following concepts through the `Rule()` function of the same name.
 
 - Name
 
     `Rule_Name = Rule()`
```
ThisIsMissEm: I'd suggest changing this to:
```suggestion
    `Rule_Name = Rule(...)`
```

To clarify that there's stuff omitted there.

docs/rules.md
```diff
@@ -54,9 +55,130 @@ My_Rule_Name_v2 = Rule(
...
+|  main.sml
+```
+
+This sort of structure lets you define rules and models that are specific to certain event types so that only the necessary rules are run for various event types. For example, you likely have some rules that should only be run on a `post` event, since only a `post` will have features like `text` or `mention_count`.
+
+Inside of each directory, you may maintain an `index.sml` file that will define the conditional logic in which the rules inside that directory are actually included for execution. Although you could handle all of this conditional logic inside of a single file, maintaining separate `index.sml`s per directory greatly helps with neat organization.
+
+## Models
+
+Before you actually write a rule, you'll need to define a "model" for an event type. For this example, we will assume that you run a social media website that lets users create posts, either at the "top level" or as a reply to another top level post. Each post may include text, mentions of other users on your network, and an optional link embed in the post. Let's say that the event's JSON structure looks like this:
```
julietshen, line 92: nit: can we say "Osprey assumes" vs "we assume" to make it super clear?

docs/rules.md
ThisIsMissEm: Or maybe structuring it slightly differently: Explain what a rule is, and then show an example, and then explain what that example does. Like:

> The general syntax for rules is `Rule_Name = Rule(parameters...)` where the Rule_Name is ... and the parameters are as follows:
> - `when_all=[ ... ]` provides the logic for the rule. [ extended description ]
> - `description=f'...'` provides the human readable description for what the logic of the rule does and why we have that rule. This can be used to help with debugging of complex workflows.

docs/rules.md
ThisIsMissEm, line 26: Does this terminate the rule evaluation? Or are all `when_all` evaluated?

docs/rules.md
```diff
@@ -54,9 +55,130 @@ My_Rule_Name_v2 = Rule(
 )
 ```
 
-## Instrumenting Rules with WhenRules
+## Rule Structuring
```
ThisIsMissEm, line 58: This is sort of covers what I was trying to ask @vinaysrao1 on the call the other day, it looks like the "Rule Storage" is by default on-disk, and follows this specific directory structure.

If the UI could create rules, then maybe you'd have some representation like this that's stored in a database?

docs/rules.md
```diff
@@ -54,9 +55,130 @@ My_Rule_Name_v2 = Rule(
...
+	path='$.user.handle',
+)
+
+PostCount: int = JsonData(
+	path='$.user.postCount',
+)
+
+AccountAgeSeconds: int = JsonData(
+	path='$.user.accountAgeSeconds',
+)
```
ThisIsMissEm, line 163: The paragraph below says:
> Here, instead of simply using `JsonData`, we instead use the `EntityJson` UDF.

but both of these are using `JsonData`, not `EntityJson` ?

docs/rules.md
```diff
@@ -54,9 +55,130 @@ My_Rule_Name_v2 = Rule(
...
+```
+
+Here, instead of simply using `JsonData`, we instead use the `EntityJson` UDF. More on this later, but as a rule of thumb, you likely will want to have values for things like a user's ID set to be entities. This will help more later, such as when doing data explorations within the Osprey UI.
+
+### Model Hierarchy
+
+In practice, you may find it useful to create a hierarchy of base models:
+
+- `base.sml` for features present in every event (user IDs, handles, account stats, etc.)
+- `account_base.sml` for features that appear only in account related events, but always appear in each account related event. Similarly, you may add one like `record_base.sml` for those features which appear in all record events.
```
ThisIsMissEm, line 173: It'd be worth defining how the loading of files happens here, does the loading look for all `*base.sml` files first, and then load up the other files?

ThisIsMissEm: i.e., what's the inheritance model here?

docs/rules.md
ThisIsMissEm:
```suggestion
        # Disabled_Rule_1,
```

I'm assuming `# ...` is a comment in the SML language, based on the two things that look like comments below.

ThisIsMissEm, line 200: If I'm understanding this correctly, when the `verdict` of the rules `Enabled_Rule_1` or `Enabled_Rule_2` is `reject` then add the following labels to the given entities, correct?

ThisIsMissEm: I guess my next question is then, can I have multiple `DeclareVerdict`s in a `then=[ ... ]` block?

e.g., do this when reject, do this when accept, etc.

docs/rules.md
```diff
@@ -79,13 +201,16 @@ WhenRules(
...
 ```
 
-WhenRules() must occur after rule creations within the file, and may become difficult to interpret outcomes of rules if too distributed so it can be beneficial to place any effects toward the bottom of workflows.
+`WhenRules()` must be placed after rule declaration within a file, and it may become difficult to interpret outcomes of rules that are too distributed. Therefore, it may be beneficial
+to place any effects toward the bottom of workflows.
 
 ## Output Sinks
 
-After the rules are all run, a set of output sinks takes the resulting `ExecutionOutput` and performs additional work based on that data. These may be defined as part of a plugin as a means to perform domain specific work.
+After all rules are evaluated for an input event, a set of output sinks takes the resulting `ExecutionOutput` and performs additional work based on that data. These may be defined
```
ThisIsMissEm: Might be worth explaining where `ExecutionOutput` comes from?

docs/rules.md
```diff
@@ -114,10 +239,11 @@ def push(self, result: ExecutionResult) -> None:
 
 ## User Defined Functions (UDFs)
```
ThisIsMissEm, line 240: I forget the markdown syntax for naming header links, but whatever that is, it could be used here and then on line 138, you could link to this section?

docs/rules.md
```diff
@@ -114,10 +239,11 @@ def push(self, result: ExecutionResult) -> None:
...
 
-User Defined Functions (UDFs) are plugins that enable users of Osprey to extend and customize their use of the Osprey SML. UDFs are implemented python functions defined and registered as a plugin. They extend the `UDFBase` abstract base class with a set of arguments, and an output. These will be executed whenever called in the sml.
+User Defined Functions (UDFs) are plugins that enable users of Osprey to extend and customize their use of the Osprey SML. UDFs are implemented as Python functions and are registered
```
ThisIsMissEm:
```suggestion
User Defined Functions (UDFs) are plugins written in Python that enable users of Osprey to extend and customize their use of the Osprey SML. UDFs are implemented as Python functions and are registered
```

docs/rules.md
```diff
@@ -220,7 +347,8 @@ Should_Warn_User_Of_Spammer = Rule(
 )
 ```
 
-Labels will also be shown in the UI for entities, and can also be set manually. Note that since the UI only searches across actions, `HasLabel` will not work in the Query UI. Instead, you may use `DidAddLabel`, which will be true when the given action added a label to a specific entity.
+Labels will also be shown in the UI for entities, and can also be set manually. Note that since the UI only searches across actions, `HasLabel()` will not work in the Query UI.
```
ThisIsMissEm, line 350: Should this be considered a bug/feature to implement support for?

docs/rules.md
```diff
@@ -252,7 +380,7 @@ MyThirdRule = Rule(when_all=[
 ])
 ```
 
-## Workflow Structure and File Placement
+### Workflow Structure and File Placement
```
ThisIsMissEm, line 383: Linking to this section from Rule Structuring might be helpful

docs/rules.md
```diff
@@ -279,3 +407,86 @@ Require(rule=f'actions/{ActionName}.sml')  # will execute 'actions/foo.sml'
...
+        ListLength(list=MentionIds) >= 1, # if there is at least one mention in the post
+    ],
+    description='First post for user includes a link embed',
+)
+
+# Finally, set which effect UDFs will be triggered
+WhenRules(
+    rules_any=[FirstPostLinkRule],
+    then=[
+        ReportRecord(
```
ThisIsMissEm, line 444: It may be worth adding a comment above this saying "This is a custom UDF that we've defined"

docs/rules.md
```diff
@@ -54,9 +55,130 @@ My_Rule_Name_v2 = Rule(
...
+|  |  |  post.sml
+|  |  |  like.sml
+|  |  account/
+|  |  |  signup.sml
+|  main.sml
+```
+
+This sort of structure lets you define rules and models that are specific to certain event types so that only the necessary rules are run for various event types. For example, you likely have some rules that should only be run on a `post` event, since only a `post` will have features like `text` or `mention_count`.
+
+Inside of each directory, you may maintain an `index.sml` file that will define the conditional logic in which the rules inside that directory are actually included for execution. Although you could handle all of this conditional logic inside of a single file, maintaining separate `index.sml`s per directory greatly helps with neat organization.
```
ThisIsMissEm: Perhaps explain here that `main.sml` in `example-rules` is the "entrypoint" and which has `Import` and `Require` statements to control which other files are loaded and when, allowing you to compose together logic.

ThisIsMissEm (commented):
This looks good, apologies about the earlier individual comments.. I thought I just had one or two comments, but I had a lot more 😅

julietshen (approved):
Looks good to me - I think overall I'm mixed on whether we use first-person pronouns or just use "Osprey" instead of "we" to help make it clear that Osprey is a self-hosted tool but defer to @ThisIsMissEm and @cassidyjames on whether that's necessary. Thank you so much for these docs!

docs/rules.md
ThisIsMissEm: Perhaps: 
> In this example, the assumption is that you run a social media website that has users who create posts. Posts can either be a reply to another post, or a top-level post. Each post may ...

cassidyjames: Here "we" is the docs author not really the product though. 😅 I agree we should try to be clear about Osprey being self-hosted, but this specifically seems like a style thing more than anything. I'm happy to merge with this and then we (ha) can always revisit the writing style over all for consistency.

docs/rules.md
haileyok: there's not actually a specific directory structure you need to use, this is just something that i came up with 😅. rules being present on disk also isn't a hard requirement - at least, that meaning you don't need them to be present in the shipped container. there's a way of getting the rules pulled in via etcd (and deployed later on via etcd too).

docs/rules.md
haileyok: great question, i am fully unsure.

docs/rules.md
haileyok: i'll go with the former for now. there should be some more stuff done on these docs later i think...

docs/rules.md
haileyok: Not actually sure, I've never used that before.

docs/rules.md
haileyok: I don't think it's a bug, this is intended. Query UDFs have to have the needed logic to rewrite SML into Druid/SQL/etc. It's probably something that could get implemented though I'm not quite sure what the query would need to be. There's a few query UDFs in the code (like `RegexMatch`) that show a good example of this though.

docs/rules.md
haileyok: tbh this needs a whole separate piece of documentation. the output sinks are definitely something most people will run into and rn there's not much info

docs/rules.md
haileyok: it's a good question that i do not know the answer to 😅

docs/rules.md
ThisIsMissEm: Oh, so this is only relevant for... development? Or is there a better way we can explain this, like "Rules can be stored on disk, or loaded from a source" or something?

docs/rules.md
ThisIsMissEm: Ha, okay! 😅

docs/rules.md
ThisIsMissEm: I can't even find `ExecutionOutput` in this repository using github search :/

docs/rules.md
ThisIsMissEm: Maybe there's an issue to implement this?

docs/rules.md
haileyok: It's probably something that should get described somewhere. I wouldn't say it's only for development - if you feel like just rebuilding a docker container each time you add new rules, that's totally fine and probably convenient for most folks. If you're working w/ multiple people though who are writing rules, its probably easier to have an etcd setup. I think both should be documented somewhere.

haileyok: i think ultimately what probably needs to happen is this single piece of documentation get split off into multiple docs pages with more info on each of these points, because they are each quite important!!

docs/rules.md
haileyok: (btw, i do think we should have some documentation for all of these stdlib UDFs and that would be a good place to explain all of these...i imagine we could get part way there w/ claude, though my experience is that claude halucinates a lot with osprey documentation so we'll have to be careful and actually make sure what gets put out is accurate)

docs/rules.md
haileyok: oh good callout. that was a typo (idk if i introduced that or not, oops) but should be `ExecutionResult` (like the code below suggests...)

docs/rules.md
haileyok: ya i have personal preferences on these things but aligning on what we want to use throughout would be solid (and honestly im so bad at maintaining consistency with this...)

PR #148, opened 2026-02-25
fix: declare worker dependencies + migrate ddtrace v2→v4

zzstoatzz:
## Motivation

We're building a plugin that depends on `osprey-worker` and ran into two issues:

1. **`pip install osprey-worker` fails** — the package only declares `flask-cors` and `osprey_rpc` as dependencies. Everything else lives in the root workspace `[dependency-groups]`, so standalone installation outside the monorepo is broken. On Python 3.14, the install pulls 10 packages instead of ~50 and immediately fails on `from osprey.engine.udf.base import UDFBase` with `ModuleNotFoundError: No module named 'typing_inspect'`.

2. **ddtrace v2 doesn't build on Python 3.14** — `ddtrace==2.21.11` (and the `<3.0.0` cap in `override-dependencies`) blocks resolution on 3.14 entirely. ddtrace v4 adds 3.14 support but changes several APIs.

## Changes

**Declare worker dependencies**: added the full runtime dependency list to `osprey_worker/pyproject.toml` (grouped by subsystem) so the package is installable outside the monorepo.

**Migrate ddtrace v2 → v4**: updated all imports in `osprey_worker/src/` — module moves (`ddtrace.span` → `ddtrace.trace`, `Pin` relocation), removed APIs (`span.sampled`), and tightened types (`set_tag` value narrowed to `str | None`).

**Bump root pins**: updated `ddtrace` (2.21.11 → 4.5.0), `envier` (0.5.2 → 0.6.1), `bytecode` (0.16.2 → 0.17.0), and widened `override-dependencies` to `>=4.0.0`. These are the minimum transitive dep bumps needed to keep the lockfile consistent — all other pins, workspace config, and `[tool.uv.sources]` are untouched.

## Verification

- `uv sync` resolves and installs cleanly on Python 3.11 (no regression from existing setup)
- standalone `pip install osprey-worker` succeeds on Python 3.14 with all core engine imports working
- `ruff check`, `ruff format --check`, and `mypy` pass on all changed files

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -25,25 +25,25 @@ members = [
     "osprey-rpc",
     "osprey-worker",
 ]
-overrides = [{ name = "ddtrace", specifier = ">=2.17.2,<3.0.0" }]
+overrides = [{ name = "ddtrace", specifier = ">=4.0.0" }]
 
 [manifest.dependency-groups]
 common = [
     { name = "absl-py", specifier = "==2.3.1" },
     { name = "asttokens", specifier = "==3.0.0" },
     { name = "blinker", specifier = "==1.9.0" },
-    { name = "bytecode", specifier = "==0.16.2" },
+    { name = "bytecode", specifier = "==0.17.0" },
     { name = "cachetools", specifier = "==5.5.2" },
     { name = "certifi", specifier = "==2025.8.3" },
     { name = "charset-normalizer", specifier = "==3.4.2" },
     { name = "click", specifier = "==7.1.2" },
     { name = "datadog", specifier = "==0.51.0" },
-    { name = "ddtrace", specifier = "==2.21.11" },
+    { name = "ddtrace", specifier = "==4.5.0" },
     { name = "decorator", specifier = "==5.2.1" },
     { name = "deepmerge", specifier = "==0.3.0" },
     { name = "dnspython", specifier = "==2.0.0" },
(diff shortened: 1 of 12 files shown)


uv.lock
```diff
@@ -1630,14 +1625,71 @@ name = "osprey-worker"
...
+    { name = "ddtrace" },
+    { name = "deepmerge" },
+    { name = "dnspython" },
+    { name = "flask" },
     { name = "flask-cors" },
+    { name = "gevent" },
+    { name = "graphviz" },
+    { name = "greenlet" },
+    { name = "grpcio", version = "1.49.1", source = { registry = "https://pypi.org/simple" }, marker = "platform_machine == 'x86_64'" },
+    { name = "grpcio", version = "1.53.2", source = { registry = "https://pypi.org/simple" }, marker = "platform_machine != 'x86_64'" },
```
ayubun, line 1639: hm, why is this added twice?

interestingly, grpcio >1.49 is known to effectively "break" osprey in production, which is very scary

i would not be surprised if this compatibility pr also has similar issues, but its hard to say

haileyok: from this i think https://github.com/roostorg/osprey/pull/141

context is that latest uv versions on silicon macs dont work for building osprey with grpicio 1.49, so doing any local dev isnt really possible :/ tho i do agree it probably breaks osprey if u were really trying to use it. maybe there's a better way to fix that tho

Issue #157, opened 2026-03-04
Updating pkg versions and confirming pkg use


annebdh:
A handful of package dependencies are running versions with known issues and need to be updated.

In looking at that list a bit more, I came across multiple dependencies where the only reference to them in the codebase was a lock/toml file (eg uv.lock or package-lock.json). It made me wonder:
1) Are we pinning transitive dependencies? (If so, do we _want_ to be doing that?)
2) Given the iterations on this codebase, are we accidentally carrying around dependencies we're no longer using?

Option 3) User error and I just couldn't find the references is also very valid :) but wanted to open this Issue to confirm before just jumping right into updates. 

Packages I was playing with as examples:
- pillow
- dnspython
- jinja2
- minimatch
- qs

julietshen:
findings: 

there are no dead direct dependencies, and nothing actually undeclared.

FawltyDeps found 3 undeclared deps:
- `flask-cors`: imported by ui_api/osprey/app.py
- `aiodns` / `pycares`: imported by `async_worker/.../async_mx_lookup.py`

BUT they are declared in their member pyprojects. fawltydeps  isn't configured to read those. Fix is to add the member pyprojects to fawltydeps' deps list (and then we can safely turn on --check-undeclared).

The ones originally listed in the issue are transitive deps or declared and used

PR #159, opened 2026-03-04
Add FawltyDeps for unused deps, single source of truth for dependencies, and turn on CI

juanmrad:
### Dependency cleanup & unused dependency detection

**FawltyDeps setup:**
Added [FawltyDeps](https://github.com/tweag/FawltyDeps) to detect unused dependencies. It runs via `uv tool run` in CI (inside the `python-quality` job after `uv sync --dev`), using the full venv for accurate package-to-module mapping.

- Configured in `[tool.fawltydeps]` to scan `osprey_worker/src`, `osprey_rpc/src`, `example_plugins/src`
- Type stubs, dev CLI tools, and gRPC server extensions are in `ignore_unused` (legitimately not imported directly)
- Ruff catches unused *imports* while FawltyDeps catches unused *dependencies*

**Cleaned up `[dependency-groups].common`:**
Removed ~60 transitive dependencies that were explicitly pinned but are already handled by the lockfile (`uv.lock`). Only direct production dependencies and type stubs remain in `common`.

- Removed transitive deps: `certifi`, `charset-normalizer`, `idna`, `urllib3` (from `requests`), `markupsafe`, `itsdangerous` (from `flask`), `zope-event`, `zope-interface` (from `gevent`), `bytecode`, `envier`, `wrapt` (from `ddtrace`), and many others
- Removed genuinely unused packages: `pillow`, `jinja2`, `imagehash`, `pdqhash`, `numpy`, `scipy`, `pywavelets`, `python-jose`, `jslog4kube`, `maxminddb`, `retry`, `rich`, `textual`, `tabulate`, `xmltodict`, and others
- Moved dev/build tools from `common` to `dev`: `mypy-extensions`, `mypy-protobuf`, `grpcio-tools`, `setuptools`, `ipdb`, `ipython`, `memray`, `pytest-flask`, `pytest-order`, `requests-mock`, `sqlalchemy2-stubs`

**Fixed `grpcio-tools` platform markers:**
Added `platform_machine` markers to `grpcio-tools` (matching the existing `grpcio` pattern) so it uses `1.49.1` on x86_64 and `1.53.*` on aarch64.

**Fixed `grpcio-health-checking` not being installed:**
It was accidentally merged onto the same line as another dependency entry and was silently ignored.

**PR template:**
Added PR template with checklist items for `uv run ruff check .`, `uv tool run fawltydeps --check-unused --pyenv .venv`.

Part of #157

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,10 @@ common = [
     "grpcio==1.53.*; platform_machine != 'x86_64'",
     "grpcio-health-checking==1.44.0",
     "grpcio-reflection==1.44.0",
-    "grpcio-status==1.49.1",
+    "grpcio-status==1.49.1; platform_machine == 'x86_64'",
+    "grpcio-status==1.53.*; platform_machine != 'x86_64'",
+    "grpcio-tools==1.49.1; platform_machine == 'x86_64'",
+    "grpcio-tools==1.53.*; platform_machine != 'x86_64'",
     "gunicorn",
     "hash-ring",
     "intervals==0.9.2",
@@ -97,8 +100,6 @@ common = [
     "types-werkzeug==1.0.9",
 ]
 dev = [
-    "grpcio-tools==1.49.1; platform_machine == 'x86_64'",
-    "grpcio-tools==1.53.*; platform_machine != 'x86_64'",
     "ipdb==0.13.13",
     "ipython==9.4.0",
     "memray==1.17.2",
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,7 @@
 
 # Maintainers, plus ROOST-maintained teams of other trusted contributors
 # See https://roostorg.github.io/community/roles.html#maintainers
-* @EXBreder @haileyok @vinaysrao1 @roostorg/roosters @roostorg/discord
+* @ayubun @EXBreder @haileyok @vinaysrao1 @roostorg/roosters @roostorg/discord
 
 # Docs contributors
 docs/ @roostorg/roosters
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "uv"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "grpcio"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,4 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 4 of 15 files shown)


pyproject.toml
```diff
@@ -38,113 +28,56 @@ common = [
...
-    "google-crc32c==1.7.1",
     "google-resumable-media==2.7.2",
     "googleapis-common-protos==1.70.0",
     "graphviz==0.20",
     "greenlet",
     "grpc-google-iam-v1==0.14.2",
-    "grpcio==1.49.1; platform_machine == 'x86_64'",  # Normal dependency for production
-    "grpcio==1.53.*; platform_machine == 'aarch64'", # Only install on ARM64 architecture for local development on MacOS
+    "grpcio==1.49.1; platform_machine == 'x86_64'",
+    "grpcio==1.53.*; platform_machine == 'aarch64'",
```
vinaysrao1: Should we use platform_machine != 'x86_64'" for the second line rather than hardwiring aarch64 (which I believe is Linux on arm.

pyproject.toml
```diff
@@ -38,113 +28,56 @@ common = [
...
     "grpcio-health-checking==1.44.0",
     "grpcio-reflection==1.44.0",
     "grpcio-status==1.49.1",
```
vinaysrao1: Does this need a platform_machine specification like line 37?

pyproject.toml
```diff
@@ -162,25 +95,24 @@ common = [
...
-    "werkzeug==1.0.1",
-    "wrapt==1.17.2",
-    "xmltodict==0.14.2",
-    "zipp==3.23.0",
-    "zope-event==5.1.1",
-    "zope-interface==7.2",
 ]
 dev = [
+    "grpcio-tools==1.49.1; platform_machine == 'x86_64'",
+    "grpcio-tools==1.53.*; platform_machine == 'aarch64'",
```
vinaysrao1: Like above should the second line be platform_machine != 'x86_64'

pyproject.toml
juanmrad: apparently not.

pyproject.toml
juanmrad: Sure. Done :)

pyproject.toml
juanmrad: Done as well.

pyproject.toml
juanmrad: Just checked and all 3 do not need platform markers as they are python code instead of native code like the grpcio. Which needs it depending on the platform the version available to use is there.

juanmrad: nvm scratch that. we do given the main library does diverge so we need markers for it to prevent new code using things the old one doesn't have 😓. I kinda dislike this. we should check if people are using this on a `x86_64` machine or not and just get rid of it. most people would be using linux I think.

PR #171, opened 2026-03-07
feat: add postgres result store

serendipty01:
Added for ease in local development

Need to change below env to enable this
OSPREY_EXECUTION_RESULT_STORAGE_BACKEND=postgres

Note:
I haven't tested with druid setup
Also, I'm not deeply familiar with Python or this codebase, so let me know anything that needs to be changed or improved.

Tested with Hailey's fork
image
image

### DIFF
--- a/osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
+++ b/osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
@@ -5,7 +5,7 @@ import json
 from abc import ABC, abstractmethod
 from datetime import datetime
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, cast
 
 import gevent
 import google.cloud.storage as storage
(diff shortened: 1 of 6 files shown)


osprey_worker/src/osprey/worker/lib/storage/stored_execution_result.py
```diff
@@ -471,6 +472,78 @@ def _execution_result_dict_from_minio_data(data: Dict[str, Any]) -> Dict[str, An
         return execution_result_dict
 
 
+class StoredExecutionResultPostgres(ExecutionResultStore):
```
ayubun, line 475: i know you're just following conventions that have existed before you, but i'm not in love with this naming scheme. perhaps for a different pr, it could be renamed to something more descriptive:
```suggestion
class PostgresExecutionResultStore(BaseExecutionResultStore):
```
the term `StoredExecutionResult` implies that the class is a pydantic object / dataclass, but this is just an interface for the result store

ayubun (approved):
i like the pr, but i don't love some of the naming in it. i think we should change all of the existing names (perhaps in a different pr)

@EXBreder @haileyok maybe we should move discord to this store for localdev, since we already have a postgres locally for other elements of the ui & the bigtable simulator is annoying to use 😆

haileyok (approved):
nice! im curious since you were testing w/ my fork and atproto data, were you handling the full firehose throughput? im curious what the perf on inserts was for you!

i pushed a changelog entry too for you, we need to put that in the pr template 😅

PR #173, opened 2026-03-07
add dependabot.yml, stop dependabot from bumping grpcio

haileyok:
Adds the dependabot.yml file with `grpcio` as an ignored dependency so it doesn't get bumped.


vinaysrao1 (approved):
Checked the post 1.48 versions of grpcio, concerns about performance and memory leaks go back to 2023.

PR #183, opened 2026-03-11
update maintainers

ayubun:
since i worked with osprey a lot within my previous team, i'd value being a maintainer, even outside of my responsibilities at discord. i care about open source software and want the best for this codebase ^^

this is a followup from the discussion in the roost discord


cassidyjames (approved):
+1 from me! @haileyok also gave an explicit +1 on Discord, and @EXBreder was the one who brought it up so I'd count that as an implicit +1 :)

PR #193, opened 2026-03-25
Use _is_spammy_exception for action_health metric filtering

cmttt:
## Summary

Fixes the `osprey.action_health` metric (added in #191) to use `_is_spammy_exception` for filtering errors instead of just `ExpectedUdfException`.

## Problem

The previous filter counted `MissingJsonPath`, `TypeError`, and `NodeFailurePropagationException` as "unexpected" errors. These are operationally expected — `MissingJsonPath` fires for every optional `JsonData` field not present in the payload, and `NodeFailurePropagationException` cascades from those. This caused ~100% "unexpected error" rate for action types like `payment_blocked` that import models with many optional fields (~211 errors per action, all from cascade).

## Changes

- Uses `_is_spammy_exception` (same filter as the existing `udf_execution` metric) to only count truly actionable errors
- Renames tag from `had_unexpected_errors` → `had_actionable_errors`
- Drops the always-true `had_errors` tag

## Test plan

- [ ] Run full test suite via Docker (`./run-tests.sh`)
- [ ] Deploy and verify `had_actionable_errors` tag appears in Datadog
- [ ] Confirm `payment_blocked` no longer shows ~100% error rate

### DIFF
--- a/osprey_worker/src/osprey/engine/executor/executor.py
+++ b/osprey_worker/src/osprey/engine/executor/executor.py
@@ -408,9 +408,7 @@ def execute(
 
     effects = context.get_effects()
 
-    actionable_error_infos = [
-        error_info for error_info in error_infos if not _is_spammy_exception(error_info.error)
-    ]
+    actionable_error_infos = [error_info for error_info in error_infos if not _is_spammy_exception(error_info.error)]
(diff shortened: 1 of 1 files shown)


haileyok (approved):
lgtm

PR #194, opened 2026-03-26
Default to selecting all for event stream

### LINKED ISSUE #126: Default to selecting all for event stream
Right now when you load the dashboard, nothing is selected. We should probably default to selecting everything so it doesn't look empty out of the box.

chimosky:
## Description

All features are now selected by default, the existing behavior still works as expected.

I did this because for some reason, it seems setting defaultChecked to true on each checkBox doesn't work.

Fixes #126

@cassidyjames kindly review.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/osprey_ui/src/components/event_stream/FeatureSelectModal.tsx
+++ b/osprey_ui/src/components/event_stream/FeatureSelectModal.tsx
@@ -76,6 +76,12 @@ const FeatureSelectModal = () => {
     return featureCategory.substring(0, featureCategory.lastIndexOf('.')).replace(/_|\//g, ' ');
   };
 
+  const isSetSelected = (feature: string) => {
+    if (!useCustomFeatures) return true;
+
+    return selectedFeatures.has(feature as string);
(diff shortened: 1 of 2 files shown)


osprey_ui/src/components/event_stream/FeatureSelectModal.tsx
```diff
@@ -76,6 +76,12 @@ const FeatureSelectModal = () => {
     return featureCategory.substring(0, featureCategory.lastIndexOf('.')).replace(/_|\//g, ' ');
   };
 
+  const setSelected = (feature: string) => {
```
haileyok: can we rename this to `isSetSelected` or similar? `setSelected` reads like a React state setter 😅

chimosky: Done, I'm new to writing Typescript so that didn't occur to me.

haileyok (approved):

PR #198, opened 2026-03-28
Add uv lock freeze command

chimosky:
## Description

uv.lock shouldn't change with every change, this makes that clear.

## Checklist

- [x] Tests pass locally

@haileyok kindly review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🎉 New features
     - Add Postgres execution result store ([#171](https://github.com/roostorg/osprey/pull/171) by [@serendipty01](https://github.com/serendipty01))
+    - Add `ParseInt` UDF — converts a numeric string to an integer ([#190](https://github.com/roostorg/osprey/pull/190) by [@bealsbe](https://github.com/bealsbe))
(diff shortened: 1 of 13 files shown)


docs/development/workflow.md
```diff
@@ -70,6 +70,10 @@ refactor: simplify rule evaluation logic
 3. **Run quality checks:**
 
    ```bash
+   # This prevents uv.lock from being modified
+   # this is preferred.
+   uv sync --frozen
+
```
haileyok, line 41: i wonder if we should just make this a different section (maybe the first thing you should run?) since i believe this would be needed even for type hinting

docs/development/workflow.md
```diff
@@ -17,25 +17,48 @@ Every commit automatically runs:
 3. **YAML/JSON/TOML validation**
 4. **Ruff linting and formatting**
 
-### Manual Checks
+## Making Changes
 
-Before pushing, run:
+1. **Prevent uv.lock modification**
```
haileyok: nit: I'd just name this "Sync dependencies". it isn't _strictly_ to avoid those `uv.lock` changes, but rather to make sure you've actually installed all the needed deps.

chimosky: Done.

haileyok: ty!

haileyok:
oh i can't merge this, cc @cassidyjames

Issue #188, opened 2026-03-23
Remove druid-specific data model


vinaysrao1:
**Is your feature request related to a problem? Please describe.**
Step one in abstracting the Osprey codebase away from assuming Druid downstream
This was an action item I took in the Osprey Working Group Meeting on Mar 20, 2026

**What solution would you like?**

osprey.engine) currently produces druid-specific filter JSON directly. This couples the rule execution core to druid. This issue is to make the filter intermediate representation to be db neutral. 

Key changes expected:
1. QueryUdfBase.to_druid_query() to a generic filter method
2. ast_druid_translator.py to generic query translator
3. Build frozen data classes for the query language operations 
4. Osprey.engine will not need pydruid - becomes optional

**Describe alternatives you have considered**
Leave druid in place. 

**Do you have any additional context?**

Issue #51, opened 2025-11-10
pluggable timeseries storage and ast translator


haileyok:
**Is your feature request related to a problem? Please describe.**

Running druid is painful to set up ([see compose ](https://github.com/roostorg/osprey/blob/main/docker-compose.yaml)) and painful to maintain. Right now however, the only way to really replace this is to maintain a forked version where you:

- Add a new client for whatever storage you want to use
- Create a new ast translator for the UI to create the SQL (or whatever) for your storage
- Create a new query model for the UI to run said queries
- Update [events.py](https://github.com/roostorg/osprey/blob/main/osprey_worker/src/osprey/worker/ui_api/osprey/views/events.py) to use this new class for marshaling queries
- Whatever is needed to actually store execution results, wherever that may live (didn't do this myself since I store outputs outside of Osprey). Ex. would be gracefully managing schema updates as new fields are added in rules (BigQuery for example doesn't support - afaict - dynamic schemas, so there's a little bit of hacking required there)

Basically, would be good to be able to just add the new client etc. to your plugins and register them at runtime. The primary piece right now that gets in the way is probably that change needed in `events.py`, though there's probably more needed as well for the actual result storage.

**What solution would you like?**

Make the storage and ast translator pluggable so that we can swap in things like Clickhouse, TimescaleDB, or BigQuery.

**Describe alternatives you have considered**

Cry.

Even short term just getting something in there instead of Druid would be great, but from what I can tell the amount of work required probably doesn't change much if its just made pluggable.

**Do you have any additional context?**

[example sql ast translator](https://gist.github.com/haileyok/b3afde0ed7a6a3cd8874819097d122c1)
[example query model](https://gist.github.com/haileyok/e48dae74e4d880a7a42ab7563e06d71c)
[example bigquery schema updating](https://gist.github.com/haileyok/e55058055e69a767d09837412e92c822)

haileyok:
I don't think its a "small" change, but it also probably isn't crazy difficult. Might be a week or t wo for working through stuff, since the plugin system is already in place. But could easily be a bit more depending on how hard it is to make all of the different pieces needed play well with plugins (see below)

There's a few design questions though that I can think of off the top of my head, but there's probably other ones that would need to get talked through

- Right now query UDFs have a `to_druid_query` method (or something similarly named, can't remember exactly) which returns a...Druid query. That means that query UDFs would only work for Druid at the moment, and we'd need to figure out an elegant way of having query UDFs support _multiple_ providers.
[...]

PR #214, opened 2026-04-09
POC clickhouse implementation

shirkevich:
## Summary

References #51 and #188.

This PR introduces a pluggable event-query backend abstraction and a ClickHouse reference implementation alongside the existing Druid path.

The main change is a backend-neutral filter IR and shared event-query contract. Druid now renders through the IR, and ClickHouse implements the same API shape for scan, timeseries, approximate group count, top-N, and CSV export.

## Why

This makes backend work explicit instead of Druid-specific by default.

It also improves local development and backend evaluation:
- ClickHouse dev mode is materially smaller than Druid in local Docker footprint
- ClickHouse reaches a queryable local stack faster in local testing
- the backend boundary is now clear enough to add other providers without rewriting the event API surface
- on macOS/Apple Silicon, the local Druid integration stack still pays for `linux/amd64` emulation, while ClickHouse does not

## What changed

- add backend-neutral filter IR and AST -> IR translation
- keep Druid compatibility by rendering Druid filters from the IR
- add `EventQueryBackend` abstraction and plugin hook
- add built-in ClickHouse backend, client holder, and feature mapper
- add ClickHouse local compose overlay and isolated ClickHouse integration test stack
- add isolated Druid integration test stack for backend-specific regression coverage
- split fast default tests from backend-specific integration suites
- add request-path validation for backend query errors so invalid feature references return `400`
- keep execution-result storage explicit and separate from the event-query backend contract
- fix ClickHouse typed-feature comparisons so missing JSON keys do not match default-value filters
- fix ClickHouse scan pagination so repeated timestamps page stably with an `action_id` tie-breaker
- reject non-zero ClickHouse top-N precision instead of silently returning different semantics
- add ClickHouse and Druid integration coverage for explicit-false vs missing-feature behavior
- wait for Druid supervisor readiness before Druid integration tests publish data
- preserve the shared Postgres DB during Druid integration runs so the Druid metadata store is not torn down mid-stack
- disable Datadog instrumentation telemetry in the test runner to avoid noisy shutdown exceptions
- make Postgres schema creation stable across repeated `make test` runs in the same Docker test stack

## Validation

- `make clickhouse-test`
- `make druid-test PYTEST_ARGS='osprey_worker/src/osprey/worker/ui_api/osprey/views/tests/test_events_druid_integration.py'`
- `make test`
- `make test` again without reset

## Local timing notes

ClickHouse remains materially faster and simpler to iterate on locally than the Druid stack.

The cold-start timing picture on local macOS/Apple Silicon is still uneven enough that I do not want to overstate a single Druid number in the PR body. The more important point for this POC is that the backend-specific suites stay isolated behind dedicated `make` targets, while the default `make test` path remains fast and repeatable.

## Rationale

This is a POC for a pluggable backend model, with ClickHouse as the first non-Druid reference implementation.

The immediate value is simpler local setup and a more practical path for production installations that do not want to carry a full Druid stack just to support the event query surface.

## Follow-ups / known gaps

- backend instance caching is still a follow-up; backend selection currently constructs a backend per request
- ClickHouse still participates only as the event-query backend; full execution-result storage remains a separate contract
- the Druid metadata DB should eventually be isolated from the app test DB instead of relying on the Druid-specific preserve behavior


cassidyjames:
@shirkevich thanks so much for your work on this!

We discussed this PR a bit in today's working group meeting #256. The consensus was that so far this looks a bit too ClickHouse-specific, and overall the series of tasks for making things pluggable should probably be broken down a bit more into smaller chunks/more iterative PRs. I know we have a bunch of related issues opened:

- #232 
  - #51 
    - #188 
    - #226 

It'd probably be most helpful next to chat with other contributors like @haileyok to determine the specific steps/how things can be broken down into iterative PRs. That's probably best done on #51, in the [#osprey Discord channel](https://discord.gg/AwxZ9K6MsU), or in the next working group (two weeks from today).

PR #221, opened 2026-04-13
add back jslog4kube dep

haileyok:
## Description

It looks like we removed this dependency in https://github.com/roostorg/osprey/pull/159/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L72, but unfortunately that breaks the osprey-ui-api from running properly.

## Checklist

- [x] Tests pass locally
- [x] Updated `CHANGELOG.md` with my changes, if applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -213,6 +213,8 @@ ignore_unused = [
     "setuptools",
     # Runtime CLI: started via command line, not imported in source
     "gunicorn",
+    # Gunicorn logger class: loaded via --logger-class flag, not imported in source
+    "jslog4kube",
     # Optional runtime dep of sentry-sdk for FlaskIntegration
     "blinker",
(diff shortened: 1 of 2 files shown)


chimosky:
I noticed this while implementing the `confluent-kafka` switch and it didn't occur to me that it was removed, because it's not a direct dependency and I expected uv to handle installing transitive dependencies, but it seems I was wrong.

julietshen (approved):
LGTM, great catch

haileyok:
i need to figure out how to get that test to pass, i'll do that later today

edit: nvm that was easy

chimosky:
I'm wondering if there's a better way to handle this rather than ignore it.

haileyok:
ya im not sure. there probably is, though since we are already doing this for other deps i think its fine for now.

PR #238, opened 2026-05-01
[osprey-ui] Migrate Antd 4 to Antd 5 (Phase 1 of UI 2.0)

haileyok:
## Summary

First bit of work to move some UI changes upstream, authored by Claude. Trying to take the easiest path here that also gets us some other free wins. Diff is large but the changes are actually quite small, majority of the diff is just lockfile.

1. Upgrade antd. There's a dark theme that got added, and to make that work with the old antd version there was some 1.5k line CSS file (lol) that Claude wrote, I assume. It's better for us to just upgrade because we get all of those changes for free from the upgrade.
2. Remove `moment` and replace it with `dayjs`. This is a migration that most folks have done, and I think its worth getting it done here too.

Claude took the liberty of setting up some basic functionality with dayjs and loading the necessary modules we need from it. It also updated all the various places we were using both moment and old-api antd to use the new version/packages.

## Changes Made (Claude)

**Dep delta** (`osprey_ui/package.json`):
- `+antd ^5.0.0` (was `4.24.16`)
- `+@ant-design/icons ^5.0.0` (was `4.8.3`)
- `+dayjs ^1.11.0` (new)
- `−moment 2.24.0`, `−moment-timezone 0.5.31`, `−@types/moment-timezone 0.5.13`

**Antd 5 wiring:**
- Wrapped `AppRouter` in `<ConfigProvider theme={{ token: { colorPrimary: '#1227ce' } }}>` + `<App as AntdApp>` so future code can use the `App.useApp()` modal/message/notification context (`osprey_ui/src/App.tsx`).
- Migrated both `Modal.error()` static calls in `BulkActionStartModal.tsx` to `App.useApp().modal.error()` (AC1.4 / AC1.7). Remaining `message.*` / `notification.*` static callsites in CopyLinkButton, SavedQueryBar, BulkLabelDrawer are intentionally left for now (out of Phase 1 scope).

**moment → dayjs:**
- New `osprey_ui/src/utils/DayjsSetup.tsx` registers the `utc`, `timezone`, `duration`, and `customParseFormat` plugins once at app entry.
- Migrated 9 files (`QueryDatePicker`, `DateUtils`, `QueryUtils`, `SearchParamUtils`, `Timeseries`, `QueryCards`, `LabelForm`, `SavedQueryCard`, `QueryTypes`).
- Dropped the legacy `as unknown as RangePickerProps['value']` cast in `QueryDatePicker.tsx` (Antd 5's RangePicker is natively typed for dayjs).
- Typed `DURATIONS` in `DateUtils.tsx` with `ManipulateType` to avoid an `as any` cast.

**Codemod + cleanup:**
- Ran `@ant-design/codemod-v5` over `osprey_ui/src/` (handles `visible` → `open`, `dropdownClassName` → `popupClassName`, etc.). Removed the global `import 'antd/dist/antd.css'` (AC1.6).
- Renamed the wrapper-interface field `visible` → `open` on `BulkActionStartModal` so the wrapper API matches Antd 5 conventions (the codemod handled the JSX, but missed the local prop interface).
- Hoisted `OspreyButton`'s `ButtonProps` import to root `'antd'` with `import type` (was `'antd/lib/button'` runtime import).
- `OspreyButton` now extends `Omit<ButtonProps, 'color'>` to keep the custom CSS-class-based `color` prop compatible with Antd 5's strict `color` union.

## Testing (Claude)
**Automated (locally green):**
- `npx tsc --noEmit` — clean
- `npm run format:check` — clean
- `npm run build` — succeeds

## Testing (Me)
Smoke tested the UI myself by going through the various screens and stuff locally. Nothing noticeable stands out as having changed from this PR to what is in main right now.


ThisIsMissEm (approved):
Looks fine to me, but I'm not super knowledgeable of this codebase.

thisandthat2:
@ThisIsMissEm  are you kidding me? i wouldn't hit this pr with johnsons dick

hilarious how terrible you all are :popcorn:

thisandthat2:
i cant stop laughing

osprey_ui/src/utils/DateUtils.tsx
```diff
@@ -1,33 +1,32 @@
...
 export function isTimestampPast(timestamp: string): boolean {
-  return moment.utc(timestamp).isBefore(moment.utc());
+  return dayjs.utc(timestamp).isBefore(dayjs.utc());
 }
 
 export function getIntervalFromDateRange({ start, end }: { start: string; end: string }): MomentRangeValues | null {
-  const momentDuration = moment.duration(moment(start).diff(moment(end)));
-  let unit: moment.unitOfTime.Base | null = null;
+  const dayjsDuration = dayjs.duration(dayjs(start).diff(dayjs(end)));
+  let unit: string | null = null;
```
juanmrad: Moving to string here removes time-unit safety we had with moment. We could use `ManipulateType` here
```suggestion
  let unit: ManipulateType | null = null;
```

osprey_ui/src/App.tsx
```diff
@@ -32,50 +32,57 @@ const AppRouter: React.FC = () => {
     updateApplicationConfig(appConfig);
   });
 
+  const brandPrimary =
+    getComputedStyle(document.documentElement).getPropertyValue('--brand-primary').trim() || '#1227ce';
```
juanmrad:
```suggestion
  const theme = React.useMemo(
    () => ({
      token: {
        colorPrimary:
          getComputedStyle(document.documentElement).getPropertyValue('--brand-primary').trim() || '#1227ce',
      },
    }),
    [],
  );
```
and update the JSX below to `<ConfigProvider theme={theme}>`.

We want to memoize here so `getComputedStyle` doesn't run on every render and, more importantly, so we don't hand a new `theme` reference to `ConfigProvider` each time. `AppRouter` re-renders whenever `useApplicationConfigStore` updates or `usePromiseResult` resolves; without the memo each of those triggers a full Antd-tree re-render for no reason.

juanmrad (approved):
two minor comments. Otherwise LGTM.

osprey_ui/src/App.tsx
haileyok: Done in 826b369. Wrapped the theme literal in a `useMemo([])` and updated the JSX to `<ConfigProvider theme={themeConfig}>` (named `themeConfig` to avoid shadowing Antd's `theme` import that phase-3 reaches for via `theme.darkAlgorithm`). Cascaded the rebase through phase-1.5/2/3/4.

PR #240, opened 2026-05-01
[osprey-ui] Persistent sidebar navigation (Phase 2 of UI 2.0)

haileyok:
## Summary

Updating the sidebar to replicate what we've moved toward internally. Screenshots attached.

## Testing (Claude)
**Automated (locally green):**
- `npx tsc --noEmit` — clean
- `npm run format:check` — clean
- `npm run build` — succeeds

## Testing (Human)

Running locally, works...


haileyok:
(Claude) Re: the inline comment on `NavBar.tsx` (`useEffect` → `toggleSidebar`) — done in `b2857ec`. Moved the localStorage write into the `toggleSidebar` callback and dropped the `useEffect`. Side note: there's now no localStorage write on mount, only on user toggles — which I think is the cleaner semantic anyway. Couldn't reply in-thread because your review is still pending and we share a GitHub identity (gh API rejected as 'one pending review per user per PR').

haileyok:
sidebar-with-logo-collapsed
sidebar-with-logo-expanded

osprey_ui/src/components/navigation/NavBar.tsx
```diff
@@ -1,93 +1,154 @@
...
 import Logo from '../../assets/Logo';
-import useApplicationConfigStore from '../../stores/ApplicationConfigStore';
-
 import { Routes } from '../../Constants';
 import styles from './NavBar.module.css';
 
-const NavBar = ({ children }: React.PropsWithChildren) => {
-  const [showMenu, setShowMenu] = React.useState(false);
-  const location = useLocation();
-  const isRecordingClicks = useApplicationConfigStore((state) => state.isRecordingClicks);
```
juanmrad, line 16: should we remove this? from my high level reading the previous NavBar applied an `isRecording` class to `.appWrapper` based on `isRecordingClicks`. 

I believe this is to prevent shift click errors from extended selection. we may want to add this back?

osprey_ui/src/components/navigation/NavBar.tsx
```diff
@@ -1,93 +1,154 @@
...
+    // Match the menu item whose key is the longest prefix of pathname.
+    const candidateKeys = [
+      Routes.HOME,
+      Routes.QUERY_HISTORY,
+      Routes.SAVED_QUERIES,
+      Routes.RULES_VISUALIZER,
+      Routes.DOCS_UDFS,
+      Routes.BULK_ACTION,
+      Routes.BULK_JOB_HISTORY,
+    ];
```
juanmrad: This is the same key set already declared one block up inside `menuItems`. Adding a new top-level route to menuItems will silently fail to highlight in the sidebar unless the contributor remembers to also add it here. Worth deriving instead.

we could do something like:
```ts
const candidateKeys: string[] = menuItems.flatMap((group) =>
  group && 'children' in group && group.children
    ? group.children.map((child) => String(child!.key))
    : [],
);
```
Then `selectedKeys` just closes over the `candidateKeys` and only needs to depend on `location.pathname`

osprey_ui/src/components/navigation/NavBar.tsx
```diff
@@ -1,93 +1,154 @@
...
+  const menuItems: MenuProps['items'] = [
+    {
+      key: 'investigate',
+      type: 'group',
+      label: 'Investigate',
+      children: [
+        {
+          key: Routes.HOME,
+          icon: <SearchOutlined />,
+          label: <Link to={Routes.HOME}>Query</Link>,
```
juanmrad: we could add some a11y here on the links by doing:
```ts
label: (
  <Link
    to={Routes.HOME}
    aria-current={location.pathname === Routes.HOME ? 'page' : undefined}
  >
    Query
  </Link>
),
```

to add the `aria-current` if it is the current page.

juanmrad (approved):
Two suggestions, and one one question.  feel free to merge once addressed :)

osprey_ui/src/components/navigation/NavBar.tsx
haileyok: TIL about `aria-current`, nice!

PR #243, opened 2026-05-04
playwright-mcp: install + register Playwright MCP devDep

haileyok:
## Summary

Since we are making a bunch of UI changes, let's add Playwright both for agent use in development as well as UI testing (eventually). Not actually implementing any tests here since that's a chunky bit of work and things are going to drift a bit, but this will set us up.

## Testing

Had Claude attempt to access `http://localhost:5002` via `browser_navigate` and used `browser_snapshot`, `browser_evaluate`, and `browser_take_screenshot`, all of which worked. Claude was able to see the basic error we provide when the UI API is not running.


AGENTS.md
```diff
@@ -89,6 +89,21 @@ cargo clippy -- -D warnings   # advisory
...
+## Browser MCP (UI verification)
+
+`osprey_ui/` registers `@playwright/mcp` as a project-scoped MCP server (`.mcp.json` at repo root). When Claude Code launches in this repo it gets `browser_navigate`, `browser_snapshot`, `browser_evaluate`, `browser_take_screenshot`, and the rest of the `browser_*` tool surface — useful for verifying visual UI changes against the running dev server without a full automated test suite. There is intentionally no `playwright.config.ts` / `@playwright/test` integration; the MCP is for ad-hoc verification, not CI.
+
+Prereqs (one-time per machine):
+
+```bash
+cd osprey_ui && node node_modules/playwright/cli.js install chromium
+sudo node osprey_ui/node_modules/playwright/cli.js install-deps chromium  # Debian/Ubuntu fresh install only
+```
```
ThisIsMissEm, line 104: I wouldn't recommend having a reference to sudo here, perhaps: "Ask the operator to install the playwright dependencies with these commands, the agent must not execute these commands itself:"

haileyok: oh yea good call. you'd hope it wouldn't do this on its own but with automode i honestly dont know what it would do

ThisIsMissEm: I've definitely heard of the agent thinking it needs sudo and trying to get it _somehow_. I explicitly have a section in my claude.md that says "hey, my git commits are signed with a key protected by touch ID, if I'm not at my computer, the commit will timeout, just pause and print the instructions for doing the commit." — I think sudo is another case where you need explicitness about the agent not actually attempting it, but rather that these are install-permissions.

ThisIsMissEm (commented):
Something I've also done in my own project is to have a `playwright-credentials.json` file (gitignored) with an example copy, and it uses the database seeders to setup users to login with (when using username/password). This then gives claude something it can reference to get a login session (if needed). That might be more important for Coop than Osprey though, not sure.

haileyok:
> and it uses the database seeders to setup users to login with (when using username/password)

yea at this point we don't have anything like this in osprey, so not very useful. we might want to have it do something for test data though, but will need to play more with that later as well once we get to having it do tests and stuff.

i iterated a bit on the prompting here in agents.md, think it looks better...linux is a lot more annoying to get this working, whereas it "just works" on macos and windows lol

.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "mcpServers": {
+        "playwright": {
+            "command": "pnpm",
```
juanmrad: Small nit, but this will require contributors to install pnpm to use the MCP, and given the project is otherwise npm based that's likely a non-trivial number of people 🤔. I do think we may want to consider migrating to pnpm as a project.

juanmrad (approved):
👍🏻

.mcp.json
haileyok: hrm yea. we should probably migrate off, but will just switch this to npx for now.

juanmrad: sounds good.

ThisIsMissEm: The risk here is that npx will allow claude to execute commands without respecting minimum release age requirements, which is the primary way malware is currently spreading in the npm registry. Usually it's detected within 24-72 hours, which pnpm can default to refuse installation or usage of packages within that window. Without that safeguard, claude could infect your machine with malware and actually execute it, should playwright get hacked.

haileyok: with the version pinned though, is this a concern?

ThisIsMissEm: Could be fine, but `npm i` has a habit of automatically updating to the latest version for packages, so, be careful.

haileyok: gotcha, thanks! I do agree we should be restricting packages that are < X days old from being used though, and i think that's probably something we should do with a migration to pnpm...

Issue #131, opened 2026-01-31
Docs: add troubleshooting section


julietshen:
Add a quickstart troubleshooting section to README covering:

- How to reset Druid schema
- Time range configuration in UI (sometimes test data doesn't show up in the UI if the time range is too large)

cassidyjames:
Fixed by #247

PR #247, opened 2026-05-11
Docs: add quickstart troubleshooting section to README

VINODvoid:
## Description

  Adds a Quickstart Troubleshooting section to the README covering the two issues raised in #131:

  - How to reset Druid schema — soft reset (terminate Kafka supervisor + restart spec-submitter) and hard reset (remove Druid/MinIO volumes, with an option to wipe Postgres too)
  - Why test data may not appear in the UI when the time range is too large, and how to fix it

  ## Checklist

  - [x] Tests pass locally
  - [ ] Updated `CHANGELOG.md` with my changes, if applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,39 @@ This is a working system, not a prototype. Try it locally, connect your data, wr
 
 Your experimentation feedback will directly shape future priorities and help us build the most useful Trust & Safety tooling for the community.
 
+## Quickstart Troubleshooting
+
+### Druid schema not updating
+
+If Druid is not ingesting data or is stuck on a stale schema after changing rule output structure, reset the Kafka supervisor and resubmit the ingestion spec:
+
+```bash
+curl -X POST http://localhost:8888/druid/indexer/v1/supervisor/osprey.execution_results/terminate
+docker compose restart druid-spec-submitter
+```
+
+To wipe all Druid and MinIO state and start fresh (Postgres data is preserved):
+
+```bash
+docker compose down
+docker volume rm osprey_middle_var osprey_historical_var osprey_broker_var osprey_coordinator_var osprey_router_var osprey_druid_shared osprey_minio_data
+docker compose up -d
+```
+
+To wipe everything including Postgres:
+
+```bash
+docker compose down -v && docker compose up -d
+```
+
+### Test data not appearing in the UI
+
+The UI defaults to querying the last 24 hours. If the selected time range is too large (weeks or months), Druid scans across many segments and results can be slow or appear empty.
+
+- narrow the time range to 1–4 hours centered on when you generated test data
+- click the edit icon next to the displayed time range to switch to a custom date/time picker
+- note that Druid's Kafka consumer uses `auto.offset.reset: latest` — it only picks up events produced after `docker compose up` first ran, so events from before that point will not appear regardless of the time range
+
 ## Recognition
 
 Discord uses Osprey to quickly detect and remove new types of harm that put users at risk. Rather than leaving other platforms to build similar tools from scratch, ROOST and Discord have open-sourced this powerful rule engine in collaboration with [internet.dev](https://internet.dev/) to make it available for anyone who needs it.
--- a/README.md
+++ b/README.md
@@ -53,39 +53,6 @@ This is a working system, not a prototype. Try it locally, connect your data, wr
 
 Your experimentation feedback will directly shape future priorities and help us build the most useful Trust & Safety tooling for the community.
 
-## Quickstart Troubleshooting
-
-### Druid schema not updating
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,39 @@ This is a working system, not a prototype. Try it locally, connect your data, wr
...
+
+To wipe everything including Postgres:
+
+```bash
+docker compose down -v && docker compose up -d
+```
+
+### Test data not appearing in the UI
+
+The UI defaults to querying the last 24 hours. If the selected time range is too large (weeks or months), Druid scans across many segments and results can be slow or appear empty.
```
haileyok: honestly we should fix this, i feel like its a bug...

cassidyjames:
Thanks for the contribution! @julietshen or @haileyok if you're able, could you give this a once-over content-wise? I'm happy with adding the section in general. 😄

README.md
```diff
@@ -53,6 +53,39 @@ This is a working system, not a prototype. Try it locally, connect your data, wr
 
 Your experimentation feedback will directly shape future priorities and help us build the most useful Trust & Safety tooling for the community.
 
+## Quickstart Troubleshooting
+
+### Druid schema not updating
+
+If Druid is not ingesting data or is stuck on a stale schema after changing rule output structure, reset the Kafka supervisor and resubmit the ingestion spec:
```
haileyok: if this is happening it would be great to try and figure out why and if we can fix it! im not aware of this being a problem, though i've also never dealt with druid here...

README.md
```diff
@@ -53,6 +53,39 @@ This is a working system, not a prototype. Try it locally, connect your data, wr
...
+docker compose down -v && docker compose up -d
+```
+
+### Test data not appearing in the UI
+
+The UI defaults to querying the last 24 hours. If the selected time range is too large (weeks or months), Druid scans across many segments and results can be slow or appear empty.
+
+- narrow the time range to 1–4 hours centered on when you generated test data
+- click the edit icon next to the displayed time range to switch to a custom date/time picker
+- note that Druid's Kafka consumer uses `auto.offset.reset: latest` — it only picks up events produced after `docker compose up` first ran, so events from before that point will not appear regardless of the time range
```
haileyok: minor nit

```suggestion
- Narrow the time range to 1–4 hours centered on when you generated test data
- Click the edit icon next to the displayed time range to switch to a custom date/time picker
- Note that Druid's Kafka consumer uses `auto.offset.reset: latest` — it only picks up events produced after `docker compose up` first ran, so events from before that point will not appear regardless of the time range
```

haileyok (commented):
thanks! im broadly fine with these suggestions (though i do feel we should be fixing these problems as well and not leaving them as hanging fruit...)

my main nit though is i think these should be in one of the places inside of `docs/` and not in the main readme (there isn't really any other development/operational info in that file and i think we should keep docs like this together)

README.md
julietshen: totally agree, and i think the default event stream view is also still buggy....

julietshen:
> thanks! im broadly fine with these suggestions (though i do feel we should be fixing these problems as well and not leaving them as hanging fruit...)
> 
> my main nit though is i think these should be in one of the places inside of `docs/` and not in the main readme (there isn't really any other development/operational info in that file and i think we should keep docs like this together)

Maybe this can go in DEVELOPMENT.md @VINODvoid ? 

FYI I'm proposing a reorg of the docs as well to address some feedback about gaps and organization. I'd love to get this in before that.

haileyok:
yup yup, i dont have any strong preferences on where it goes right now either since we need to do a big pass over the docs for cleanup anyway!

VINODvoid:
Thanks for the thorough review @haileyok and @julietshen  !

**on file placement** : Makes sense - I'will move this into docs/ (DEVELOPMENT.md as @julietshen  suuggested). Since you want this before the reorg, I will update the PR now and let you know when it's ready for another look.
 
**on fixing root causes**: Agreed these shouldn't just be hanging fruit. Should I open issues for both (Druid schema + UI empty data behaviour ) so they are tracked separately. This PR can document the workarounds for now while the fixes are investigated.

Let me know if there's a preferred spot inside docs/ beyond DEVELOPMENT.md

VINODvoid (commented):
Done — moved the troubleshooting content from README.md into docs/DEVELOPMENT.md and applied @haileyok's capitalization suggestion. Let me know if there's a more specific spot inside docs/ you'd prefer, @julietshen.

haileyok (approved):
thank you!!

haileyok:
> should I open issues for both (Druid schema + UI empty data behaviour ) so they are tracked separately.

yes please!! there might be an open issue for the empty UI data behavior already...though i definitely don't know about the druid schema problem, so that might be a fun one to get a small issue on!

PR #249, opened 2026-05-12
fix: add retention limits to Kafka topics to prevent unbounded disk growth

VINODvoid:
Fixes #239. Topics `osprey.actions_input` and `osprey.execution_results` are now
  created with 48-hour / 8 GB-per-partition retention and 1 GB segments. Also
  documents the `kafka-configs` command for existing deployments.

  ## Description

  - Added `--config retention.ms`, `--config retention.bytes`, and `--config
  segment.bytes`
    to both `kafka-topics --create` commands in `docker-compose.yaml`
  - Added a Troubleshooting section to `docs/DEVELOPMENT.md` with the `kafka-configs`
    command for operators on existing deployments

  ## Checklist

- [ ] Tests pass locally
- [ ] `uv run ruff check .` passes (no unused imports or other lint errors)
- [ ] `uv tool run fawltydeps --check-unused --pyenv .venv` passes (no unused dependencies)
- [ ] Updated `CHANGELOG.md` with my changes, if applicable

PR #250, opened 2026-05-13
fix: use line continuations in kafka-topics bash command

VINODvoid:
## Problem

  Without `\` line continuations, each indented flag inside `bash -c "..."` in docker-compose.yaml was parsed as a separate shell command rather than part of the `kafka-topics`
  invocation. This caused topic creation to fail silently in CI.

  ## Fix

  Add `\` at the end of each continuation line in both `kafka-topics` commands.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -143,3 +143,17 @@ docker compose --profile test_data up osprey-kafka-test-data-producer -d
 ```
 
 Produces user login events with timestamps, user IDs, and IP addresses to `osprey.actions_input` topic.
+
+## Troubleshooting
+
+### Kafka topic disk growth
+
+Topics are created with a 48-hour / 8 GB-per-partition retention limit. If you deployed before this was set, apply the config to your existing topics:
+
+```bash
+for topic in osprey.actions_input osprey.execution_results; do
+  kafka-configs --bootstrap-server localhost:9092 \
+    --entity-type topics --entity-name $topic --alter \
+    --add-config retention.ms=172800000,retention.bytes=8589934592,segment.bytes=1073741824
+done
+```
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -78,15 +78,15 @@ services:
         condition: service_healthy
     command: >
       bash -c "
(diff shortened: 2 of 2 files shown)


docs/DEVELOPMENT.md
haileyok, line 190: huh isnt this already in there? https://github.com/roostorg/osprey/blob/main/docs/DEVELOPMENT.md#kafka-topic-disk-growth (im not sure why github isn't showing this as existing already or as being a conflict...)

haileyok (approved):
a q but lgtm

docs/DEVELOPMENT.md
haileyok: oh i see there's a conflict now maybe that's what it needs resolved...

VINODvoid: only net-new change here is the \ line continuations in docker-compose — the DEVELOPMENT.md bit was already in from #249. those missing backslashes were what broke CI i guess !!!

haileyok: ahh yea, had to fix that conflict first. diff looks good now, thanks!!

haileyok:
@VINODvoid ugh sorry for this back and forth...this is super cursed lol. i think the `\` characters are broken here because YAML (sigh...)

can we just make these one line each. unfortunate because it looks ugly but will at least work...

```
kafka-topics --bootstrap-server osprey-kafka:29092 --create --if-not-exists --topic osprey.actions_input --partitions 3 --replication-factor 1 --config retention.ms=172800000 --config retenti
         +on.bytes=8589934592 --config segment.bytes=1073741824 &&
kafka-topics --bootstrap-server osprey-kafka:29092 --create --if-not-exists --topic osprey.execution_results --partitions 3 --replication-factor 1 --config retention.ms=172800000 --config ret
         +ention.bytes=8589934592 --config segment.bytes=1073741824 &&
kafka-topics --bootstrap-server osprey-kafka:29092 --list
 ```

(thank you, and sorry again...annoying because the updated action won't even run until merge lol)

VINODvoid:
@haileyok  no worries, I will try to comeup with better solution, meanwhile can u ping to some issues that requires like immediate attention

PR #275, opened 2026-05-19
fix(deps): mypy 1.13 → 1.20 with type tightening

haileyok:
Supersedes #216, which only carried the dependabot lockfile bump and would not pass CI on its own.

mypy 1.20 (a wide jump from 1.13) broke type checking in three categories. This PR carries the lockfile bump plus the fixes needed to make it land cleanly.

### 1. Plugin compatibility crashes
Both type-checking plugins crashed during `uv run mypy .`:
- `pydantic.mypy` at `pydantic/mypy.py:868` (`TypeError: tuple object expected; got list` — mypy 1.20 tightened `add_method`'s args param)
- `sqlalchemy.ext.mypy.plugin` at `sqlalchemy/ext/mypy/infer.py:405` (deeper crash in column inference)

Fixed by bumping to the latest 1.x patches on each plugin's host package:
- `pydantic` 1.10.19 → 1.10.26
- `sqlalchemy` 1.4.31 → 1.4.54

(These are minor scope expansions but necessary — without them, mypy can't run at all on this codebase against 1.20.)

### 2. Tighter `ast.Constant.value` typeshed
mypy 1.20 propagates the precise union `str | bytes | int | float | complex | EllipsisType | None` where it previously collapsed to `Any`. Five call sites in `py_ast.py` plus one each in `arguments.py` and `experiments.py` consume `ast.Str.s` / `ast.Constant.value` and now require narrowing.

Pattern used: `assert isinstance(node.value, expected_type)` matching the existing assert pattern in those files. In `experiments.py:246` (a set comprehension over heterogeneous `Literal` subclasses) I used `cast(Any, _).value` since the runtime invariant covers multiple subclasses that don't share a common typed `.value`.

### 3. Better narrowing in `validate_static_types.py`
mypy 1.20 no longer accepts that `previous_generic_kwarg` is non-None on the branch where `resolved_typevar_type is not None` (the invariant exists in the loop but isn't a syntactic narrowing). Added an explicit `assert`.

## What is intentionally suppressed

Two cases where the type system can't follow what's actually true:

- **`bulk_action_task.py`** (3 columns): the sqlalchemy 1.4 mypy plugin infers `Mapped[T]` from `Column(T)` without reading the `nullable=` argument, so `Optional[T] = Column(Text, nullable=True)` produces a spurious LHS/RHS mismatch. Suppressed with `# type: ignore[misc]` and a one-line comment naming the plugin limitation.

- **`postgres.py`**: `sessionmaker[Session]` no longer satisfies the `_TSessionMakerType` Protocol bound — `Session` doesn't structurally implement `_SessionProtocol` under mypy 1.20's stricter Protocol checking. Suppressed with `# type: ignore[type-var]`.

## Verification

- `uv run mypy .` — clean (345 source files, was failing with 8 errors + 2 plugin crashes)
- `uv run ruff check . && uv run ruff format --check .` — clean
- `uv run pytest osprey_worker/src/osprey/engine/ -m 'not use_osprey_stdlib'` — 804 passed, 1 skipped, 4 xfailed, 63 deselected (Docker-only)

The one test that errors (`test_feature_name_to_entity_type_mapping.py`) requires Docker-backed labels storage and errors on `main` too.


juanmrad (approved):
lgtm

PR #297, opened 2026-05-24
feat(event-stream): show sensible defaults so first-load isn't empty

### LINKED ISSUE #96: Friction Log for setting up Osprey
I used Claude Code to run Osprey, here's a friction log of where I got stuck and what changes I made to make things work properly:

# Osprey Setup Friction Log

This document records all issues encountered while setting up the Osprey project from scratch on December 12, 2025.

## Environment
- **OS**: macOS (Darwin 24.5.0)
- **Platform**: Apple Silicon (ARM64)
- **Python**: 3.11
- **Date**: December 12, 2025

---

### Problem
When running `docker compose up -d`, the Postgres container immediately crashed with exit code 1.

### Root Cause
Postgres 18 changed its data storage structure to use major-version-specific directory names compatible with `pg_ctlcluster`. The docker-compose.yaml was configured with the old mount point `/var/lib/postgresql/data`, and there was existing data from a previous Postgres version in the volume.

### Error Message
```
Error: in 18+, these Docker images are configured to store database data in a
       format which is compatible with "pg_ctlcluster" (specifically, using
       major-version-specific directory names).

       Counter to that, there appears to be PostgreSQL data in:
         /var/lib/postgresql/data (unused mount/volume)
```

### Solution
1. Updated `docker-compose.yaml` line 261:
   ```yaml
   # Before
   - metadata_data:/var/lib/postgresql/data

   # After
   - metadata_data:/var/lib/postgresql
   ```

2. Removed old volumes with `docker compose down -v`

### Files Modified
- `docker-compose.yaml:261`

---

### Problem
The `osprey-ui-api` container kept crashing on startup with database connection errors.

### Root Cause
The UI API service had a basic `depends_on` configuration that didn't wait for Postgres to be healthy before starting. It attempted to connect to Postgres before the database was ready to accept connections.

### Error Message
```
psycopg2.OperationalError: could not translate host name "postgres" to address: Name or service not known
```

### Solution
Updated `docker-compose.yaml` lines 144-156 to use health check conditions:
```yaml
# Before
depends_on:
  - osprey-worker
  - druid-broker
  - postgres
  - snowflake-id-worker
  - bigtable
  - bigtable-initializer

# After
depends_on:
  osprey-worker:
    condition: service_started
  druid-broker:
    condition: service_started
  postgres:
    condition: service_healthy
  snowflake-id-worker:
    condition: service_healthy
  bigtable:
    condition: service_healthy
  bigtable-initializer:
    condition: service_completed_successfully
```

### Files Modified
- `docker-compose.yaml:144-156`

---

### Problem
Events were appearing in the Event Stream panel on the right side of the UI, but each event showed "No data for selected features" instead of the actual field values.

### Root Cause
**UX Issue**: The Osprey UI requires users to explicitly select which fields they want to display in the event stream. This is not immediately obvious to new users.

### Solution
Click the **"Select Summary Features"** button (top right of Event Stream panel) and select fields to display:
- ContainsHello
- PostText
- UserId
- EventType

### Recommendation for Improvement
Consider either:
1. Pre-selecting common fields by default
2. Adding a tooltip/hint when events show "No data for selected features"
3. Auto-selecting all available fields on first use

---

### Problem
The Timeseries Chart consistently showed "No data available" even though the Event Stream showed matching events.

### Root Cause
**Time Range Mismatch**: The query time range ended at 1:44pm EST, but data with `ContainsHello` field only started being ingested at 1:50pm EST (after we fixed some conflicts we between local rule experiments and the example data that was restored). This created a 6-minute gap where no matching data existed.

### Additional Factor
**Browser Caching**: Earlier queries that returned empty results were cached, so even after data became available, the cached empty response was being shown.

### Solution
1. Adjust the end time in the date range selector to be after 1:50pm EST (current time)
2. Hard refresh the page (Cmd+Shift+R on Mac, Ctrl+Shift+R on Windows)
3. Resubmit the query

### Verification
Direct Druid query confirmed data existed:
```json
{
  "total": 189,
  "with_hello": 60
}
```

### Recommendation for Improvement
1. Default to "now" for end time instead of a fixed historical timestamp
2. Add a "Last hour" or "Last 30 minutes" quick select option
3. Show a more helpful message when no data is found (e.g., "No data in selected time range")

---

### Problem
Multiple Docker containers showed platform mismatch warnings during startup:

```
The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8)
```

### Affected Images
- `gcr.io/google.com/cloudsdktool/cloud-sdk:latest` (bigtable)
- `apache/druid:34.0.0` (all druid containers)
- `ghcr.io/ayubun/snowflake-id-worker:0`

### Impact
**Non-blocking**: The containers run successfully under emulation, but there may be performance implications.

### Recommendation for Improvement
Consider providing ARM64 native images or documenting the expected platform in the README.

---

### Files Modified
1. `docker-compose.yaml`
   - Line 261: Changed Postgres volume mount
   - Lines 144-156: Added health check conditions for osprey-ui-api dependencies

2. `druid/specs/execution_results.json`
   - Line 8: Changed offset reset from "latest" to "earliest"

### Files Restored (from git)
- `example_rules/main.sml`
- `example_plugins/src/register_plugins.py`
- `example_rules/config/labels.yaml`

---

## Time to Resolution

| Issue | Discovery | Resolution | Time Spent |
|-------|-----------|------------|------------|
| Postgres 18 mount | Immediate | 5 mins | 5 mins |
| UI API crashes | 2 mins | 3 mins | 5 mins |
| Wrong rules loaded | 10 mins | 5 mins | 15 mins |
| Druid schema | 5 mins | 10 mins | 15 mins |
| UI features | 2 mins | 1 min | 3 mins |
| Time range | 5 mins | 2 mins | 7 mins |
| **Total** | | | **~50 mins** |

---

### What Went Well
1. **Dependencies installed cleanly** - `uv sync` worked perfectly on first try
2. **Pre-commit hooks** - Installed without issues
3. **Test data generator** - Started and worked immediately once rules were fixed
4. **Documentation** - README.md was clear and accurate for the basic setup
5. **Error messages** - Most errors (especially Postgres) had clear, actionable error messages

### Infrastructure Highlights
- Docker Compose setup is well-structured
- Health checks are properly configured (once we used them)
- Druid ingestion works reliably with schema discovery
- Worker processes events correctly once rules are loaded

---

### For New Users
1. **Always run `git status`** before assuming the repository is clean
2. **Check docker logs** (`docker compose logs <service>`) when services fail
3. **Allow 30-60 seconds** after restarting services for Druid to ingest new schema
4. **Use "now" or recent timestamps** for query end times, not historical dates

### For the Project
1. **Add a quickstart troubleshooting section** to README covering:
   - Postgres volume issues on upgrade
   - How to reset Druid schema
   - Time range configuration in UI

2. **Improve onboarding UX**:
   - Pre-select common fields in Event Stream
   - Default time ranges to "last hour" instead of fixed dates
   - Add inline help for "No data for selected features"

3. **Add a setup verification script** that:
   - Checks if services are healthy
   - Verifies data is flowing through the pipeline
   - Confirms Druid has ingested recent data

4. **Document platform compatibility** for ARM64/M-series Macs

5. **Consider adding `.dockerignore`** patterns to prevent local rule changes from being mounted into containers, or add a warning in the README about uncommitted changes

haileyok:
## Summary

Fixes sub-item 3 of the friction log (roostorg/osprey#96; related to roostorg/osprey#126): events appearing in the Event Stream panel with every row showing "No data for selected features" on first load.

### Root cause

`EventStream.tsx` renders summary features per event via `defaultSummaryFeatures.filter(f => f.appliesTo(actionName)).map(f => f.features)`. When a deployment has not curated `default_summary_features` in `ui_config` (the default backend value is `[]`), this collapses to an empty array for every event, which `EventStreamCard` displays as "No data for selected features."

### Fix

Introduced a small helper `getSummaryFeaturesForEvent(event, defaultSummaryFeatures)` that:

1. **Preserves existing behavior** when the deployment has curated `default_summary_features` and one of them matches the event's `ActionName`.
2. **Falls back** to a single block containing every key on the event (minus `ActionName`, which is already shown as the card title) when no curated default applies.

This means:
- Deployments with proper `default_summary_features` config: zero behavior change.
- Users who manually picked features via "Select Summary Features": zero behavior change.
- Fresh / demo deployments: cards now render the actual fields present on each event automatically.

### Why this approach (vs. the alternatives in the dispatch prompt)

- **Approach 1** (hardcode a guaranteed default set like `UserId`/`ActionId`) — there is no schema guarantee of any specific extracted feature beyond `ActionName`, which is already in the card title. Hardcoding fields the backend may not ship would just reintroduce the empty-row problem under a different name.
- **Approach 2** (helpful empty-state) — still requires the user to click something before seeing data. The friction log's goal is "demo works on first load."
- **Approach 3** (pre-select all known features globally) — would render "No data" rows for every event that lacks any given field. Per-event scoping avoids that.

This is essentially a per-event variant of approach 3, scoped to the keys actually present on each event.

### Scope

Only `osprey_ui/src/components/event_stream/` is touched. No backend, no infra, no `FeatureSelectModal` rewrite, no date-picker code.

## Test plan

- [x] Unit tests added for `getSummaryFeaturesForEvent` covering: defaults match, multiple defaults match, no defaults configured, defaults exist but none match, only ActionName, empty event. TDD discipline verified — temporarily reverted the fallback branch and confirmed the two new-behavior tests went RED before restoring.
- [x] `tsc --noEmit` clean
- [x] `npm run format:check` clean
- [x] Existing behavior for `customSummaryFeatures != null` and for matching `defaultSummaryFeatures` is preserved (only the empty-fallback path is new).
- [ ] Live browser verification (operator) — load a fresh deployment, scan a recent window, confirm event cards render with their actual fields instead of "No data for selected features." Then set a custom feature selection and confirm the picker still overrides correctly.

Closes #96
Refs #126

## Summary by CodeRabbit

* **Bug Fixes**
  * Event stream cards now choose the most relevant summary fields automatically based on the event type.
  * When no matching summary configuration exists, the app falls back to showing available event details instead of leaving the card empty.
  * Event row heights now stay aligned with the content being displayed for more consistent scrolling.


julietshen:
Thank you for this PR! I tested locally and things seem fine, but i think this PR helped reveal a hidden bug where the UI crashes (both for selecting specific features AND scrolling quickly through the event stream). 

https://github.com/user-attachments/assets/3b353eb3-8239-473b-b08e-7ce684a35676

claude says:
_EntityWithPopover.tsx:107 has always called decodeURIComponent(entityId) unguarded. Any entity value with a stray % (or any malformed escape) throws. Before this PR, fresh demos rendered "No data for selected features" so EntityWithPopover was never invoked with most entity values. PR #297's fallback exposes far more entity values to it._  


I actually have something locally that can fix it so I'll open an issue for `safeDecode` and that way this can land cleanly

julietshen (approved):
lgtm

haileyok:
fascinating lol, love that for us. thanks for fixing that!!

PR #321, opened 2026-05-29
Bump grpcio to 1.53.2

julietshen:
## Summary

- Bumps the grpcio family from `1.49.1` → `1.53.2` to fix [GHSA-cfgp-2977-2fmm](https://github.com/grpc/grpc/security/advisories/GHSA-cfgp-2977-2fmm) (high-severity DoS via excessive iteration in gRPC).
- Aligns `grpcio`, `grpcio-status`, `grpcio-tools`, `grpcio-health-checking`, `grpcio-reflection` on a single version (`1.53.2`).
- Removes the `platform_machine` split — that was a workaround for missing aarch64 wheels at `1.49.1`; `1.53.2` ships wheels for both x86_64 and aarch64.

Closes #316. Part of the Dependabot triage tracked in #219.

## Test plan

- [x] `uv lock` + `uv sync --dev` — resolves cleanly
- [x] `uv run ruff check` — clean
- [x] `uv run ruff format --check` — clean
- [x] `uv tool run fawltydeps --check-unused --pyenv .venv` — no unused deps
- [x] Imported every `*_pb2*` module in `osprey_rpc` against the new runtime — all 18 load cleanly
- [x] Spot-checked the gRPC APIs Osprey uses: `ServicerContext`, `StatusCode`, `intercept_channel`, `insecure_channel`, `grpc_health.v1.health_pb2`, `grpc_reflection.v1alpha.reflection` — all present
- [x] Engine test suite (804 tests) passes; the 43 errors are the pre-existing `KeyError: 'POSTGRES_HOSTS'` and require the docker-compose stack — unrelated to this change
- [ ] CI integration suite (`./run-tests.sh`) — will run in CI

`osprey_coordinator` (Rust) is unaffected: it uses `tonic`/`prost`, not the Python `grpcio` runtime.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Consolidated gRPC dependency versions to 1.53.2 with uniform pinning across all packages
  * Enhanced test environment with additional gRPC debugging capabilities for improved diagnostics

Issue #304, opened 2026-05-29
Design (expanded/desired) integration tests


julietshen:
Related to https://github.com/roostorg/osprey/issues/219, we need to determine what integration tests should cover and do some high-level design before implementation.

- [ ] @julietshen to think through other scenarios from a T&S perspective.

- [ ] @cassidyjames to create sub-issues for each of the above

julietshen:
Some ideas:

## group by 6 modes of failure that impact safety work

  1. **Silent data loss:** events accepted but never evaluated, or verdicts produced but never sunk.                                                                               
  2. **Silent staleness:** rules edited but not reloaded, labels updated but not visible, Druid behind.
  3. **Silent wrong-answer:** a UDF throws, a rule errors, but the event is treated as "no match" instead of surfaced.                                                             
  4. **Silent partial success:** one effect fires (label added) but another doesn't (ban skipped), with no signal.                                                                 
  5. **Silent misroute**: schema-malformed events accepted into the pipeline without diagnostics.                                                                                  
  6. **Silent config drift:** plugin fails to register, rule file fails to import, server boots anyway.                                                                            
                                                                                                       

## Pipeline smoke (end-to-end happy path)                                                                                                                                     
                  
-  Single event → verdict → effects. Produce one create_post action to osprey.actions_input, assert: rule evaluated (visible in execution-results bucket on MinIO), effects emitted to osprey.execution_results, entity label persisted in Bigtable, label visible via the UI API /labels endpoint. (Covers modes 1, 4.)
-  Batch burst. Use example_data/generate_spam_burst.sh as the seed; produce N events, assert every event has a corresponding result row and no Kafka consumer-lag remains  after a bounded wait. (Mode 1.)  

## Rule lifecycle                                                                                                                                                             
                  
-  Invalid rule fails to load loudly. Boot the worker with a syntactically-broken .sml; assert the worker exits non-zero and the error names the offending file + line. (Mode 6.)       
- Rule referencing missing UDF fails validation. Same shape, surfaces config drift before traffic flows. (Mode 6.)                                                       
- Rule reload picks up edits. Modify a rule file mid-test, trigger reload, assert the next event uses the new rule (and an event evaluated mid-reload uses either old-or-new but is never lost). (Mode 2.)  

## UDF / plugin failure modes
-  UDF raises. Register a UDF that raises on a specific input; assert the event's execution result is marked errored, the error is in the result payload, and downstream sinks see the error rather than a silent "no match." (Mode 3.)
[...]

PR #236, opened 2026-04-30
[examples] Bluesky/ATProto sample integration (input stream, rules)

haileyok:
## Description

Adds a new workspace package, `example_atproto_plugins`, demonstrating how to plug Osprey into a real-world event source via the existing `register_input_stream` pluggy hook. The sample subscribes to Bluesky's [JetStream](https://docs.bsky.app/blog/jetstream) WebSocket firehose and maps each ATProto commit (posts, likes, reposts, follows) into an Osprey `Action`. Companion rules live under `example_atproto_rules/`.

The main motivation is making it cheap to exercise Osprey changes against realistic event shapes and volume — the existing `osprey-kafka-test-data-producer` emits one synthetic event per second from a single template, which doesn't catch issues that only show up with real production traffic. JetStream is free, public, no-auth, and runs at hundreds of events per second; in a few minutes of local testing this stack processed ~34k events of varied content.

A secondary benefit: it fills the gap between the README's "used by Bluesky" claim and the lack of any actual ATProto reference code in this repo. The existing `example_plugins` is Discord-shaped; this is the Bluesky-shaped counterpart.

### Usage

```sh
./run-atproto.sh up
```

Stacks `docker-compose.atproto.yaml` on top of the main compose file, switching the worker from `OSPREY_INPUT_STREAM_SOURCE=kafka` to `=plugin` and pointing it at `example_atproto_rules/`. The existing demo path (`./demo.sh`) is unchanged.

### Notes / non-goals

- Sample-quality, not production-quality: no durable cursor on restart, no zstd compression, no DID-level filtering. JetStream itself is documented by Bluesky as not part of the formal AT Protocol spec — fine for a sample, treat it as illustrative for adopters.
- New runtime dep: `websocket-client==1.8.0` (Apache-2.0).
- `uv.lock` revision bumps from 2 to 3 because the project's `[build-system]` requires `uv_build>=0.8.12,<0.9.0`, which expects the newer lockfile format.
- `osprey_worker/Dockerfile` picks up the new workspace member alongside `example_plugins`.

## Confidence Level

Confidence Level: Claude

## Testing

- Unit-tested the `_event_to_action` mapping for posts, likes, deletes, identity-event skip, and missing `time_us`.
- `docker compose -f docker-compose.yaml -f docker-compose.atproto.yaml config` merges cleanly; main `docker-compose.yaml` still validates standalone.
- `uv run ruff check`, `uv run ruff format --check` clean.
[...]

PR #326, opened 2026-06-01
Add osprey-stress harness for measuring pipeline drop rate + latency

### LINKED ISSUE #324: Build a stress-test CLI for input sinks + rate-of-success measurement
Sub-issue of #304 (integration test design). Proposed by @ayu during review of #321.

## Problem

Bumping dependencies in Osprey is risky because the existing test suite doesn't tell us whether the system behaves correctly under sustained load. Unit and engine tests cover correctness on individual events; the integration suite spins up the stack but doesn't pressure-test it. Several previous grpcio/protobuf bump attempts (per ayu, \"failed every time we have tried it\") would have been caught earlier by a load-driven smoke test.

Beyond dep bumps, the same tool would help triage prod alerts (\"is this drop in throughput real?\") and gate future PRs that touch the hot path.

## What this asks for

A standalone CLI under `osprey_worker/` (or its own subdir) that:

1. **Spams the input sink** with synthetic events at a configurable rate (events/sec, burst size, duration).
2. **Measures**:
   - Events accepted by the input topic (Kafka producer ack rate)
   - Events evaluated by the worker (consumer-lag drain + result-storage count)
   - Effects dispatched (label sink writes, output topic deliveries)
   - End-to-end p50 / p95 / p99 latency from produce → result-stored
3. **Reports a single rate-of-success** number plus a structured breakdown of where events dropped (input → evaluation → effects → storage).
4. **Exits non-zero** if success rate falls below a configurable threshold, so it can be run as a CI gate.

## Suggested shape

```bash
osprey-stress \\
  --events-per-second 1000 \\
  --duration 60s \\
  --threshold 0.999 \\
  --rules ./example_rules \\
  --report json
```

The synthetic event generator can start from `example_data/generate_spam_burst.sh` (which already produces realistic-shaped actions); the CLI wraps that with measurement.

## Where this fits in the #304 design

This is the concrete implementation of two of the proposed integration-test scenarios from the #304 design memo:

- **A2. Batch burst** — produce N events, assert every event has a corresponding result row and no Kafka consumer-lag remains after a bounded wait
- **I1. Spam burst from one entity** — N events from the same `user_id`, assert threshold-crossing event triggers exactly once

A general-purpose CLI satisfies both (with different config) and gives us a foundation for the other load-shaped scenarios in #304.

## Use cases

- **Dep bumps** (this PR's motivating use case): run before merging to verify the bumped runtime doesn't drop events or slow drain
- **PR gate**: opt-in CI job that runs for ~60s on PRs touching `osprey_worker/src/osprey/`
- **Release tests** (P0 in #219 SDLC checklist): part of the release-tag verification
- **Prod alert triage**: \"is this 5% drop in throughput real or measurement noise?\"

## Acceptance criteria

- [ ] CLI installs as `osprey-stress` (or similar) from `osprey_worker`'s entry points
- [ ] Configurable rate, duration, threshold
- [ ] Structured JSON report with per-stage drop counts
- [ ] Documented in `docs/DEVELOPMENT.md` with a worked example
- [ ] Optional CI workflow that runs it against the docker-compose stack on demand (label-triggered, not every PR)

Context: surfaced during review of #321 (grpcio bump). Sub-issue of #304.

julietshen:
## Summary

End-to-end stress harness for the worker pipeline. Produces N synthetic events to the input topic at rate R, observes their `ExecutionResult`s on the output topic, and reports drop rate + p50/p95/p99 round-trip latency. Exits non-zero on threshold breach so it can gate CI on pipeline health.

Closes #324. Sub-issue of #304 (integration test design).

## Why

Bumping dependencies in Osprey has historically been risky because the existing test suite doesn't tell us whether the system behaves correctly under sustained load — see [the discussion on #321](https://github.com/roostorg/osprey/pull/321) where @ayu observed that grpcio bumps "have failed every time we have tried it." The bundled `generate_test_data.sh` produces one event per second, three orders of magnitude below realistic load, so it can't surface throughput regressions either.

`osprey-stress` gives us a controlled scenario stress test: a known number of events, a known rate, a known expected drop rate, and an exit code we can gate CI on.

## Complementary to #236, not competing

PR [#236](https://github.com/roostorg/osprey/pull/236) (hailey's jetstream input-stream plugin) provides a *realistic ambient* event source at ~500 events/s of real Bluesky public traffic. This PR provides a *controlled synthetic* event source where we know every action_id we sent.

The two are complementary:

| | #236 (jetstream) | #324 (this PR) |
|---|---|---|
| Source | Real Bluesky traffic | Synthetic |
| Rate | ~500/s, variable | Configurable |
| Determinism | Non-deterministic | Reproducible |
| Per-event matching | No | Yes (via `ActionId`) |
| External dependency | Yes (Bluesky uptime) | No |
| Best for | Soak testing | Controlled scenarios |

The measurement layer (`consumer.py` + `reporter.py`) is **producer-agnostic** by design — both modules accept an optional `action_id_filter` so they can run in closed-loop mode (per-event matching, known set of IDs) or open-loop mode (count everything during a window). The CLI has a `measure` subcommand reserved for open-loop mode that will activate once #236 lands; today it prints a stub message pointing at this connection.

## Architecture

```
osprey_worker/src/osprey/worker/stress/
├── reporter.py    # pure stats: percentile, drop-rate, threshold-breach detection
├── producer.py    # threaded Kafka producer at configurable rate, deterministic ids
├── consumer.py    # threaded Kafka consumer with filter / drain / lifecycle
├── cli.py         # argparse with `run` + reserved `measure` subcommands
└── tests/         # 49 tests using injected FakeKafka{Producer,Consumer} factories
```

Plus, in the engine:
- `ExecutionContext.get_action_id()` — symmetric with `get_action_name()`
- `GetActionId` stdlib UDF — mirrors `GetActionName` exactly
- `example_rules/models/base.sml` — exposes `ActionId = GetActionId()` so the harness can round-trip-match

The engine change matters separately from the harness: it gives any rule author a way to emit `action_id` as an extracted feature, which is a normal thing to want for audit trails, debugging, or dedup. It's split into its own commit (`feb0c5b`) for reviewability.

## Usage

```bash
osprey-stress run \
  --events 10000 --rate 1000 \
  --threshold-drop-rate 0.01 \
  --threshold-p95-ms 500 \
  --report json
```

Exit codes: `0` pass, `1` drop-rate exceeded, `2` latency exceeded, `3` internal error.

## Test plan

- [x] `pytest osprey_worker/src/osprey/worker/stress/tests/` — 49 tests pass
- [x] `pytest osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_get_action_id.py` — pass
- [x] Full engine suite — 853 tests still pass; pre-existing `POSTGRES_HOSTS` env errors unrelated
- [x] `uv run mypy` — clean
- [x] `uv tool run fawltydeps --check-unused --pyenv .venv` — no unused deps
- [x] `osprey-stress --help` works after `uv sync`
- [ ] Live run against docker-compose stack — to be done locally after this lands (the harness itself *is* the integration test for this)

## What's deliberately not in this PR

- Effects-side measurement (counting label sink writes etc.) — follow-up if useful in practice
- Open-loop `measure` subcommand body — activates with #236
- CI workflow that runs this on every PR — separate proposal once we've used it manually

## Summary by CodeRabbit

* **New Features**
  * Added `osprey-stress` CLI tool that runs closed-loop stress tests, emits synthetic events, reports drop rate and latency percentiles with threshold gating, and returns exit codes on threshold breaches
  * Added `GetActionId()` standard library function to access action IDs within execution context

* **Tests**
  * Added comprehensive test coverage for stress harness components

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### 🛠 Breaking changes
 
 ### 🎉 New features
+- Add `osprey-stress` CLI: closed-loop stress harness that produces synthetic events at a configurable rate, observes their `ExecutionResult`s on the output topic, and reports drop rate + p50/p95/p99 latency. Exits non-zero on threshold breach so it can gate CI on pipeline health. Complementary to the jetstream input-stream plugin in [#236](https://github.com/roostorg/osprey/pull/236) — the same measurement layer will run in open-loop mode against jetstream traffic once that lands.
+- Add `GetActionId()` stdlib UDF + corresponding `ExecutionContext.get_action_id()` method, symmetric with the existing `GetActionName()`. `example_rules/models/base.sml` now exposes `ActionId = GetActionId()` so the stress harness can round-trip-match events without overloading other entity fields.
 - Add Postgres execution result store ([#171](https://github.com/roostorg/osprey/pull/171) by [@serendipty01](https://github.com/serendipty01))
 - Add `ParseInt` UDF — converts a numeric string to an integer ([#190](https://github.com/roostorg/osprey/pull/190) by [@bealsbe](https://github.com/bealsbe))
 - Add `StringSlice` UDF which extracts a substring by index range ([#189](https://github.com/roostorg/osprey/pull/189) by [@bealsbe](https://github.com/bealsbe))
--- a/example_rules/models/base.sml
+++ b/example_rules/models/base.sml
@@ -12,3 +12,5 @@ EventType: Entity[str] = EntityJson(
 
 ActionName=GetActionName()
 
+ActionId=GetActionId()
+
--- a/osprey_worker/pyproject.toml
+++ b/osprey_worker/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
 ]
 [project.scripts]
 osprey-cli = "osprey.worker.lib.cli:cli"
+osprey-stress = "osprey.worker.stress.cli:main"
 
 [project.entry-points."osprey_plugin"]
 stdlib = "osprey.worker._stdlibplugin"
--- a/osprey_worker/src/osprey/engine/executor/execution_context.py
+++ b/osprey_worker/src/osprey/engine/executor/execution_context.py
@@ -172,6 +172,10 @@ class ExecutionContext:
         """Returns the action name that the execution context is currently being invoked upon."""
         return self._action.action_name
 
+    def get_action_id(self) -> int:
+        """Returns the action id of the event that the execution context is currently being invoked upon."""
+        return self._action.action_id
+
     def get_action_time(self) -> datetime:
         """Returns the time of the action that the execution context is currently being invoked upon."""
         return self._action.timestamp
--- a/osprey_worker/src/osprey/engine/stdlib/udfs/get_action_id.py
+++ b/osprey_worker/src/osprey/engine/stdlib/udfs/get_action_id.py
@@ -1,0 +1,15 @@
+from ._prelude import ArgumentsBase, ExecutionContext, UDFBase
+from .categories import UdfCategories
+
+
+class Arguments(ArgumentsBase):
+    pass
+
+
+class GetActionId(UDFBase[Arguments, int]):
+    """Returns the Action ID of the event being processed."""
+
+    category = UdfCategories.ENGINE
+
+    def execute(self, execution_context: ExecutionContext, arguments: Arguments) -> int:
+        return execution_context.get_action_id()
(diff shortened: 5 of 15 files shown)


julietshen:
Closing in favor of a smaller-PR split, per AGENTS.md ("Keep diffs small and focused"). The single 1372-line PR is too big to review honestly.

Splitting into:
- A: `GetActionId` UDF + `ExecutionContext.get_action_id()` + `base.sml` (~50 lines)
- B: Stress reporter (pure stats) + tests (~340 lines)
- C: Stress producer + tests (~365 lines)
- D: Stress consumer + tests (~370 lines)
- E: CLI + entry point + CHANGELOG (~290 lines, blocked until B/C/D land)

A–D will be opened in parallel as independent PRs (each branches off main). E goes up as draft.

PR #327, opened 2026-06-01
Add GetActionId() stdlib UDF + ExecutionContext.get_action_id()

julietshen:
## Summary

Symmetric with the existing `GetActionName()` UDF: gives SML rules access to the `action_id` of the event being processed, so they can emit it as an extracted feature for round-trip tracking, audit trails, or dedup.

`example_rules/models/base.sml` is updated to expose `ActionId = GetActionId()` alongside the existing `ActionName`, so the default example rule set's output includes the action_id without requiring every rule author to wire it up themselves.

First of a five-PR split for the stress harness work tracked in #324. This PR is the engine-level primitive; the harness itself uses it but other rule authors will benefit too (audit trails, debugging, dedup, idempotency).

## Test plan

- [x] `pytest osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_get_action_id.py` — passes
- [x] `pytest osprey_worker/src/osprey/engine/stdlib/udfs/tests/test_get_action_name.py` — still passes (regression check on the existing symmetric UDF)
- [x] `uv run mypy` — clean
- [ ] Full integration suite — runs in CI

## Related

- Sibling PRs (to come): reporter, producer, consumer, CLI for `osprey-stress`
- Sub-issue of #304 (integration test design)

## Summary by CodeRabbit

* **New Features**
  * New GetActionId UDF available for rules to access the current action identifier.

* **Tests**
  * Added test coverage for GetActionId behavior.

### DIFF
--- a/example_rules/models/base.sml
+++ b/example_rules/models/base.sml
@@ -12,3 +12,5 @@ EventType: Entity[str] = EntityJson(
 
 ActionName=GetActionName()
 
+ActionId=GetActionId()
+
--- a/osprey_worker/src/osprey/engine/executor/execution_context.py
+++ b/osprey_worker/src/osprey/engine/executor/execution_context.py
@@ -172,6 +172,10 @@ class ExecutionContext:
         """Returns the action name that the execution context is currently being invoked upon."""
         return self._action.action_name
 
+    def get_action_id(self) -> int:
+        """Returns the action id of the event that the execution context is currently being invoked upon."""
+        return self._action.action_id
+
     def get_action_time(self) -> datetime:
         """Returns the time of the action that the execution context is currently being invoked upon."""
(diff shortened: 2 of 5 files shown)


haileyok (approved):

julietshen:
Addressed CodeRabbit's note in `dfe3e43` — dropped the `List[Callable[[Any], Any]]` annotation on `pytestmark` and the now-unused `typing` imports. Lint + mypy + the test itself still pass.

Left the sibling `test_get_action_name.py` carrying the same legacy pattern unchanged, to keep this diff scoped per AGENTS.md ("Keep diffs small and focused"). Easy follow-up sweep if anyone wants to normalize the rest of the udf tests.

PR #328, opened 2026-06-01
Add stress.reporter — pure stats for the upcoming stress harness

julietshen:
## Summary

Pure-function stats library that takes per-event produce/consume timestamps and emits a structured report: drop count and rate, percentile latencies, threshold-breach detection, JSON + human format, and an exit-code policy for CI gating.

Designed to be producer-agnostic so the same module serves both:
* **closed-loop** measurement (known set of action_ids, exact 1:1 matching)
* **open-loop** measurement (count anything that shows up in a time window)

No Kafka, no I/O, no live deps. Fully unit-tested in isolation.

Part 2 of 5 in the stress-harness split for #324, per AGENTS.md (\"Keep diffs small and focused\"). Other parts:
- [#327](https://github.com/roostorg/osprey/pull/327): `GetActionId` UDF
- (this PR): reporter
- (to come): producer
- (to come): consumer
- (to come): CLI + entry point

This PR is independent — does not depend on #327 or any sibling PR. The next PRs (producer, consumer, CLI) will import from this module.

## Test plan

- [x] 15 unit tests covering: percentile math edge cases, closed-loop vs open-loop discrimination, drop-rate calculation, threshold breach detection, JSON roundtrip, human format
- [x] `uv run ruff check` — clean
- [x] `uv run mypy` — clean
- [ ] CI runs the full test suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a stress-harness reporter that calculates match/drop metrics from event timestamps and computes latency distributions (min, p50, p95, p99, max).
  * Supports optional thresholds for drop rate and p95 latency, returning deterministic CLI exit codes when limits are exceeded.
  * Generates both JSON output and a human-readable summary, including closed-loop vs open-loop reporting behavior.
* **Tests**
  * Added unit tests covering percentile validation, latency aggregation, reporting modes, threshold breach detection and exit codes, and JSON/human serialization.

PR #329, opened 2026-06-01
Add stress.producer — synthetic Kafka producer for the stress harness

julietshen:
## Summary

Threaded Kafka producer that emits N well-formed Osprey actions to a configurable topic at a configurable rate. Returns the wall-clock send time per `action_id` so the reporter (#328) can later compute end-to-end latency.

Key design calls:
* **Deterministic integer action_ids** of the form `(base + run_bucket * 10M + n)`, encoding the run_id and sequence so concurrent runs don't collide and the consumer can filter on the run.
* **JS-safe integers**: all action_ids stay below 2**53 so JSON consumers (Druid, browsers) don't lose precision.
* **Rate control** via \"sleep until next slot\" with drift correction: a long stall resets the schedule to now rather than triggering a burst-catchup.
* **Factory-injected** `KafkaProducer` so unit tests run without live Kafka.

Part 3 of 5 in the stress-harness split for #324, per AGENTS.md. Sibling PRs:
- [#327](https://github.com/roostorg/osprey/pull/327): `GetActionId` UDF
- [#328](https://github.com/roostorg/osprey/pull/328): reporter
- (this PR): producer
- (to come): consumer
- (to come): CLI + entry point

Independent of all siblings.

## Test plan

- [x] 16 unit tests using an in-memory `FakeKafkaProducer`: produces exact count, action_ids unique within a run, action_ids unique across runs, timestamps recorded per event, rate control approximately honored, `stop()` aborts early, factory failure captured, run_id helper format
- [x] `uv run mypy` — clean

## Summary by CodeRabbit

* **New Features**
  * Added a stress-testing producer that publishes well-formed Osprey events to Kafka at a configurable rate, using deterministic per-run event generation and recording per-event send timestamps for end-to-end latency measurements.

* **Tests**
  * Added comprehensive unit and integration-style tests for deterministic event payloads, action-id generation (uniqueness and safe-integer constraints), rate control behavior, correct shutdown/abort behavior, and error handling (including failing producer initialization).

### DIFF
--- a/osprey_worker/src/osprey/worker/stress/producer.py
+++ b/osprey_worker/src/osprey/worker/stress/producer.py
@@ -1,0 +1,155 @@
+"""Stress harness synthetic producer.
+
+Emits N well-formed Osprey actions to the Kafka input topic at a configurable
+rate, recording the wall-clock send time per `action_id` so the reporter can
+compute end-to-end latency.
+
+Events match the shape expected by `example_rules/` (the rule set the
+test_runner is configured with). The consumer matches each input event to its
+output `ExecutionResult` via the `ActionId` extracted feature, which is
+surfaced by the `GetActionId()` stdlib UDF.
+"""
+
+from __future__ import annotations
+
+import json
+import threading
+import time
+import uuid
+from dataclasses import dataclass
+from typing import Optional
+
+from kafka import KafkaProducer
+
+# action_ids in this range sit above what the bundled `generate_test_data.sh`
+# emits but below 2**53, so they survive JSON's integer-as-float precision.
+# A per-run multiplier keyed on the low bits of the run uuid keeps concurrent
+# stress runs from colliding with each other.
+_ACTION_ID_BASE = 1_000_000_000_000
+_RUN_BUCKET = 10_000_000
+
+
+def _action_id_for(run_id: str, n: int) -> int:
+    """Deterministic, per-run-unique integer action_id.
+
+    The id encodes the run_id (so we can filter on the consumer side) and the
+    event sequence number, all without exceeding JS safe-integer precision.
+    """
+    bucket = int(run_id[:6], 16) % _RUN_BUCKET
+    return _ACTION_ID_BASE + bucket * _RUN_BUCKET + n
+
+
+@dataclass(frozen=True)
+class ProducerConfig:
+    bootstrap_servers: list[str]
+    topic: str
+    events: int
+    rate_per_second: float
+    run_id: str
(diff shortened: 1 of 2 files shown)


reitblatt:
@coderabbitai review

reitblatt (approved):
LGTM, just take the github-code-quality suggestion on the empty exception block.

julietshen:
Rebased on latest main and addressed the comments in `505c423`:

**Accepted (both in this commit):**
- **github-code-quality + @reitblatt** on `producer.py:154` — replaced the empty `except: pass` with explicit handling. Cleanup errors are now recorded in `self._error`, but only if no earlier send-loop error was already captured. Preserves "don't raise from cleanup" without silently dropping the failure.
- **CodeRabbit** on `test_producer.py:4` — replaced `Any` with `object` (kwargs we accept but don't touch) and `NoReturn` (always-raises `boom()`).

**Declining with reasoning:**
- **CodeRabbit on `producer.py:22` (add `kafka-python` to `osprey_worker/pyproject.toml`)** — repo is a [uv workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/) and `kafka-python==1.4.7` is declared once in the root `pyproject.toml`'s `[dependency-groups].common`. Existing imports from `kafka` across `_stdlibplugin/sink_register.py`, `sinks/sink/kafka_output_sink.py`, `sinks/sink/input_stream.py`, `cli/sinks.py` etc. all rely on this single declaration. Duplicating into the workspace member would create version drift over time.
- **CodeRabbit on `producer.py:53` (convert `ProducerConfig` to Pydantic `BaseModel`)** — same reasoning I gave on #328 when declining the equivalent suggestion for `LatencyStats`/`Thresholds`/`Report`: `ProducerConfig` is an internal pure-data config with no validation boundary. The codebase uses `@dataclass` for similar internal shapes (`Action`, `ExecutionResult`); pydantic is reserved for boundaries where validation/serialization matter.

16/16 producer tests still pass, ruff + mypy clean.

julietshen:
Good catch from coderabbit, fixed in `86d6b77`. The previous version still had both calls under one `try`, so `flush()` raising would skip `close()` and leak the socket. Split into a two-step loop where each call's failure is captured but doesn't prevent the next step from running. Either error gets recorded in `self._error` only if the send loop above didn't already set a more useful one.

Added a regression test (`test_close_runs_even_if_flush_raises`) with a subclassed FakeKafkaProducer whose flush() raises — pins that close_called still becomes True. 17/17 producer tests pass.

cassidyjames (approved):

PR #330, opened 2026-06-01
Add stress.consumer — Kafka result consumer for the stress harness

julietshen:
## Summary

Threaded Kafka consumer that subscribes to `osprey.execution_results` and records the wall-clock receive time per `ActionId`. Producer-agnostic by design — works for both closed-loop and open-loop measurement:

| Mode | `action_id_filter` | Behaviour |
|---|---|---|
| **Closed-loop** | set of known ids | Records only matching events; can stop early once filter is complete |
| **Open-loop** | `None` | Records every event in the window |

The open-loop mode is what lets this same module measure jetstream traffic once [#236](https://github.com/roostorg/osprey/pull/236) lands — same code, different config.

Belt-and-suspenders parsing:
* Skips malformed JSON
* Skips messages missing `ActionId`
* Skips non-integer `ActionId` (defensive against future rule schema changes)
* First-write-wins for duplicate IDs (duplicate effect dispatches don't reset the receive timestamp)

Bounded poll cycle (`consumer_timeout_ms=200`) lets `stop()` fire within ~200ms regardless of message arrival rate.

Factory-injected `KafkaConsumer` so unit tests run without live Kafka.

Part 4 of 5 in the stress-harness split for #324, per AGENTS.md. Sibling PRs:
- [#327](https://github.com/roostorg/osprey/pull/327): `GetActionId` UDF
- [#328](https://github.com/roostorg/osprey/pull/328): reporter
- [#329](https://github.com/roostorg/osprey/pull/329): producer
- (this PR): consumer
- [#367](https://github.com/roostorg/osprey/pull/367): CLI + entry point

Independent of all siblings.

## Test plan

- [x] 11 unit tests using an in-memory `FakeKafkaConsumer`: closed-loop filtering, non-matching IDs ignored, early-stop when filter complete, first-write-wins on duplicates, open-loop count-everything, malformed-JSON skipped, missing-`ActionId` skipped, non-int-`ActionId` skipped, lifecycle (start-twice raises), factory failure captured, max-runtime honored
- [x] `uv run mypy` — clean

## Summary by CodeRabbit

* **New Features**
  * Added a Kafka-backed stress result consumer that runs in the background and records first-receive timestamps per ActionId.
  * Supports optional ActionId filtering, automatic stopping on completion or timeout, and exposes runtime errors for observability.

* **Tests**
  * Added a comprehensive test suite covering filtered/unfiltered consumption, early stopping, duplicate handling (first-write-wins), malformed message tolerance, validation errors for missing/non-integer ActionId, lifecycle behavior (start once), factory failure handling, and max-runtime enforcement.

### DIFF
--- a/osprey_worker/src/osprey/worker/stress/consumer.py
+++ b/osprey_worker/src/osprey/worker/stress/consumer.py
@@ -127,10 +127,21 @@ class Consumer:
         try:
             payload = json.loads(raw.decode('utf-8'))
         except (ValueError, UnicodeDecodeError):
+            # Garbage at the byte level is tolerated — Kafka can carry malformed
+            # bytes from broken producers and we shouldn't take down the
+            # measurement run for transport noise.
             return False
         action_id = payload.get('ActionId')
         if not isinstance(action_id, int):
-            return False
+            # Schema-level violation: a well-formed JSON result missing a valid
+            # ActionId means the harness was pointed at rules that didn't wire
+            # GetActionId() (or wired it wrong). Raise loudly so the caller sees
+            # the misconfiguration via consumer.error instead of getting silent
+            # zero matches.
+            raise ValueError(
+                f'result message has missing or non-int ActionId (got {type(action_id).__name__}); '
+                f'ensure the rule set exposes ActionId via GetActionId()'
+            )
         if self._config.action_id_filter is not None and action_id not in self._config.action_id_filter:
             return False
         # First-write-wins so duplicate effect dispatches don't reset the time.
--- a/osprey_worker/src/osprey/worker/stress/consumer.py
+++ b/osprey_worker/src/osprey/worker/stress/consumer.py
@@ -1,0 +1,137 @@
+"""Stress harness result consumer.
+
+Subscribes to the `osprey.execution_results` Kafka topic and records the
+wall-clock receive time per `ActionId` so the reporter can pair every input
+event with its output and compute end-to-end latency.
+
+Producer-agnostic: the same consumer works against synthetic load (#324),
+jetstream-driven load (once #236 lands), or any other input source. The
+optional `action_id_filter` decides whether we're in closed-loop matching mode
+(only count IDs we know about) or open-loop throughput mode (count everything).
+"""
+
+from __future__ import annotations
+
+import json
+import threading
+import time
+from dataclasses import dataclass
+from typing import Optional
+
+from kafka import KafkaConsumer
+
+
+@dataclass(frozen=True)
+class ConsumerConfig:
+    bootstrap_servers: list[str]
+    topic: str
+    group_id: str
+    # If set, only record action_ids in this set. None = open-loop, count all.
+    action_id_filter: Optional[frozenset[int]] = None
+    # Max wall-clock time to keep reading after start() is called.
+    max_runtime_seconds: float = 60.0
+    # If we've matched every id in action_id_filter, stop early.
+    stop_when_filter_complete: bool = True
+    client_id: str = 'osprey-stress-consumer'
+
+
+class Consumer:
+    """Consumes execution_results on a background thread.
+
+    `start()` -> `wait()` -> read `consumed`. `stop()` to abort early.
+    """
+
+    def __init__(self, config: ConsumerConfig, *, consumer_factory=KafkaConsumer):
+        self._config = config
+        self._consumer_factory = consumer_factory
+        self._consumed: dict[int, float] = {}
+        self._stop_event = threading.Event()
+        self._thread: Optional[threading.Thread] = None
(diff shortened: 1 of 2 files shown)


julietshen:
Addressed the github-code-quality lint in `fddcb63` — added an inline comment explaining why the `except Exception: pass` around `consumer.close()` is intentional (best-effort shutdown; we don't want a noisy close() error to mask the consumed/error state the caller needs to read). 11/11 consumer tests still pass.

Note: CodeRabbit's earlier review on this PR was rate-limited and produced no actionable inline comments. Happy to re-trigger if useful.

julietshen:
Addressed CodeRabbit's two new notes:

**1. Flaky duplicate-id test (accepted, fixed in `bb35812`).** Real bug — the test was reading `consumer.consumed.get(1)` after a fixed 200ms sleep, which could return None on slow CI and silently turn the equality assertion into a no-op. Replaced the sleep with a bounded poll that returns as soon as the first record is captured, plus an explicit assert that we actually got it before proceeding. 11/11 tests pass.

**2. `consumer_factory` type annotation (declining).** CodeRabbit themselves marked this 💤 Low value. The implicit type from the `KafkaConsumer` default works at the use sites, and a precise `Callable[..., KafkaConsumer]` would over-constrain — the test passes `lambda *a, **kw: fake` where `fake` is a duck-typed in-memory stub. Same pattern lives in producer.py; keeping consistent. If we want strict factory typing later, that's a separate cleanup pass across the harness.

julietshen:
CI failure root cause: my `FakeKafkaConsumer.__next__` was raising `StopIteration` immediately on an empty queue, but a real `KafkaConsumer.__next__` blocks for up to `consumer_timeout_ms` before doing so. The Consumer under test depends on that natural backpressure — without it, my fake turned the outer poll loop into a tight spin. Locally the GIL scheduled fairly enough that fed messages still landed; CI's thread contention starved the main test thread before any `feed()` could acquire the queue lock, leaving `consumed` empty and 6/11 tests failing.

Fixed in `433b26d` by giving `FakeKafkaConsumer` an Event-driven 50ms idle poll that mirrors kafka-python's actual semantics. `feed()` sets the event so we wake immediately when a real message arrives — no fixed latency in steady state. 5/5 clean local runs.

The consumer.py production code is unchanged — it was correctly designed; only the test fake's contract was wrong.

osprey_worker/src/osprey/worker/stress/consumer.py
```diff
@@ -0,0 +1,140 @@
...
+                pass
+
+    def _record(self, raw: bytes) -> bool:
+        """Parse one result message; return True if it counted toward consumed."""
+        try:
+            payload = json.loads(raw.decode('utf-8'))
+        except (ValueError, UnicodeDecodeError):
+            return False
+        action_id = payload.get('ActionId')
+        if not isinstance(action_id, int):
```
reitblatt, line 135: Isn't this an error case, suggesting something broken? Should we throw a ValueError?

reitblatt (commented):
Other than throwing an exception in the invalid action_id case, LGTM, but I've never used Kafka before, so take that with a grain of salt.

julietshen:
fixed issues flagged by @reitblatt and rebased on latest main (the previous `integration-tests CANCELLED` was from staleness).

osprey_worker/src/osprey/worker/stress/consumer.py
julietshen: thank you!! addressed in `a6e1f6a`.

Replaced the silent-skip on missing/non-int ActionId with a clear `ValueError`. The consumer thread's outer try/except captures it into `self.error`, so the CLI surfaces it instead of users getting confusing zero-match runs.

Two distinct failure modes, only one of which raises:
- **Malformed JSON at the byte level** — still tolerated. Kafka can carry garbage from broken producers; that's transport noise, not a contract violation.
- **Well-formed JSON missing/wrong-typed ActionId** — now raises. This means the rule set didn't wire `GetActionId()` correctly, which is a real configuration error worth surfacing.

Updated the two tests (`test_skips_missing_action_id_field` → `test_raises_on_missing_action_id_field`, same shape for non-int) to assert the new error path. 11/11 consumer tests still pass, ruff + mypy clean.

cassidyjames (approved):

julietshen:
hmm something is up with the docker-compose part of the integration test. The tests pass but the teardown isn't working. 

`
Error: The operation was canceled.  
  Terminate orphan process: pid (5968) (run-tests.sh)
  Terminate orphan process: pid (5969) (docker)
  Terminate orphan process: pid (5985) (docker-compose)
`

@cassidyjames I'm going to merge this since the tests did pass, and I'll open another issue to figure out what's going on with the github actions

PR #337, opened 2026-06-04
Add pluggable LLM API provider hook with tool calling

haileyok:
## What

Adds a new, vendor-neutral **LLM API provider plugin hook** to Osprey, designed for **tool calling from day one**.

This is the first chunk of making LLM access pluggable in Osprey. It is deliberately upstream-only and adds a new *public* plugin hook, so I'd love interface feedback before this lands (opening as draft for that reason). Relates to the Jan 9 2026 working-group "pluginizing" discussion (#104).

## Changes

- **`osprey.worker.lib.llm`** — new `BaseLLMProvider` interface plus small vendor-neutral dataclasses: `LLMMessage`, `ToolDefinition`, `ToolCall`, `ToolResult`, `LLMResponse`, `LLMUsage`, `CacheControl`. Tool definitions go in; tool-call requests come back out; tool results are fed back in on the next `chat()` call. Streaming is intentionally out of scope (room left for a future `chat_stream`).
- **`register_llm_provider` hookspec** (`firstresult=True`) + **`bootstrap_llm_provider(config)`** helper, mirroring the existing single-provider hooks (`register_input_stream`, `register_execution_result_store`). `bootstrap_llm_provider` returns `None` when no plugin registers one, so callers null-check.
- **Example provider** in `example_plugins` (`llm/anthropic_provider.py`) that talks to the **Anthropic Messages API** directly, including `tool_use` request/response translation and optional prompt-cache control. Registered via `register_llm_provider` in `example_plugins/src/register_plugins.py`.
- **Tests**: dataclass invariants + interface contract (`osprey_worker/.../lib/llm/tests/test_base.py`), and a full tool-call cycle against a fake Anthropic client — no SDK, API key, or network (`example_plugins/src/llm/tests/test_anthropic_provider.py`). `example_plugins/src` is added to `pytest` `testpaths` so example tests run.
- **Docs**: hook menu table + an "LLM provider hook" section in `docs/DEVELOPMENT.md`.

## Design notes / open questions

- **Tool calling shape.** The dataclasses are a proposal; happy to refine names/shape. The key constraint was making tool calling first-class rather than bolted on later.
- **`CacheControl` generality.** Included because a downstream consumer wants to keep Anthropic prompt-cache breakpoints (ephemeral + TTL). It's optional and vendor-neutral; if folks prefer, it could move into provider-specific `**params` instead.
- **`firstresult=True`.** Only one LLM provider can register, matching `register_input_stream`. Reasonable for now; noted in the docstring.

## Validation

Locally green: `ruff check`, `ruff format --diff`, `mypy .` (no issues), full `pre-commit run --all-files` (prettier skipped, as in CI), and `fawltydeps --check-unused`. The example provider tests pass directly; the `lib/llm` tests run under the docker integration harness like the rest of `osprey_worker`.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -125,6 +125,52 @@ def register_ast_validators() -> None:
     # Register AST validators
 ```
 
+### Available hooks
+
+Implement any subset of these in your plugin's `register_plugins.py`:
+
+| Hook | Returns | Notes |
+| --- | --- | --- |
+| `register_udfs` | `Sequence[Type[UDFBase]]` | Custom user-defined functions. |
+| `register_output_sinks` | `Sequence[BaseOutputSink]` | Where execution results go. |
+| `register_ast_validators` | `Sequence[Type[BaseValidator]]` | Extra SML validators. |
+| `register_action_proto_deserializer` | `ActionProtoDeserializer \| None` | Custom action proto → JSON. |
+| `register_input_stream` | `BaseInputStream` | Single-provider (`firstresult`). |
+| `register_execution_result_store` | `ExecutionResultStore` | Single-provider (`firstresult`). |
+| `register_labels_service_or_provider` | `LabelsServiceBase \| LabelsProvider` | Single-provider (`firstresult`). |
+| `register_llm_provider` | `BaseLLMProvider` | Single-provider (`firstresult`). LLM API access for AI-assisted features. |
+
+### LLM provider hook
+
+`register_llm_provider` lets a plugin supply the LLM API client used by AI-assisted
+features (e.g. natural-language query building). The interface lives in
+`osprey.worker.lib.llm` and is vendor-neutral and **tool-calling aware**: you pass
+`ToolDefinition`s in, the model may return `ToolCall`s, and you feed `ToolResult`s
+back on the next `chat()` call.
+
+```python
+from osprey.worker.lib.config import Config
+from osprey.worker.lib.llm.base import BaseLLMProvider
+
+@hookimpl_osprey
+def register_llm_provider(config: Config) -> BaseLLMProvider:
+    return MyLLMProvider(config)
+```
+
+Only one provider may be registered (`firstresult=True`). Retrieve it with
+`bootstrap_llm_provider(config)` from `osprey.worker.adaptor.plugin_manager`, which
+returns `None` when no plugin registers one — so callers should null-check.
+
+A direct Anthropic implementation is provided as a reference in
+`example_plugins/src/llm/anthropic_provider.py`, including the request/response and
+`tool_use` translation. It imports the `anthropic` SDK lazily; the SDK is **not**
+declared as a workspace dependency (it conflicts with the pinned
+`typing-extensions`), so install it manually to actually run the provider:
+
+```bash
+uv pip install anthropic
+```
+
 ## Rules
 
 Rules are written in SML, some examples are provided in `example_rules/` with YAML config, the rules are mounted to the worker processes when the containers start via environment variables. ex:
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -143,7 +143,7 @@ Implement any subset of these in your plugin's `register_plugins.py`:
 ### LLM provider hook
 
 `register_llm_provider` lets a plugin supply the LLM API client used by AI-assisted
-features (e.g. natural-language query building). The interface lives in
+features such as natural-language query building. The interface lives in
 `osprey.worker.lib.llm` and is vendor-neutral and **tool-calling aware**: you pass
 `ToolDefinition`s in, the model may return `ToolCall`s, and you feed `ToolResult`s
 back on the next `chat()` call.
@@ -163,13 +163,10 @@ returns `None` when no plugin registers one — so callers should null-check.
 
 A direct Anthropic implementation is provided as a reference in
 `example_plugins/src/llm/anthropic_provider.py`, including the request/response and
-`tool_use` translation. It imports the `anthropic` SDK lazily; the SDK is **not**
-declared as a workspace dependency (it conflicts with the pinned
-`typing-extensions`), so install it manually to actually run the provider:
-
-```bash
-uv pip install anthropic
-```
+`tool_use` translation. The `anthropic` SDK is a dependency of `example_plugins`
+(installed by `uv sync`); set `ANTHROPIC_API_KEY` (or the
+`OSPREY_LLM_ANTHROPIC_API_KEY` config key) to use it. The example plugins do **not**
+register it by default — add your own `register_llm_provider` hookimpl to enable it.
 
 ## Rules
 
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -140,66 +140,9 @@ Implement any subset of these in your plugin's `register_plugins.py`:
 | `register_labels_service_or_provider` | `LabelsServiceBase \| LabelsProvider` | Single-provider (`firstresult`). |
 | `register_llm_provider` | `BaseLLMProvider` | Single-provider (`firstresult`). LLM API access for AI-assisted features. |
 
-### LLM provider hook
-
-`register_llm_provider` lets a plugin supply the LLM API client used by AI-assisted
-features such as natural-language query building. The interface lives in
-`osprey.worker.lib.llm` and is vendor-neutral and **tool-calling aware**: you pass
-`ToolDefinition`s in, the model may return `ToolCall`s, and you feed `ToolResult`s
-back on the next `chat()` call.
-
-```python
-from osprey.worker.lib.config import Config
-from osprey.worker.lib.llm.base import BaseLLMProvider
-
-@hookimpl_osprey
-def register_llm_provider(config: Config) -> BaseLLMProvider:
-    return MyLLMProvider(config)
-```
-
-Only one provider may be registered (`firstresult=True`). Retrieve it with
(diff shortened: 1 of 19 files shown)


docs/DEVELOPMENT.md
```diff
@@ -125,6 +125,52 @@ def register_ast_validators() -> None:
...
+| `register_action_proto_deserializer` | `ActionProtoDeserializer \| None` | Custom action proto → JSON. |
+| `register_input_stream` | `BaseInputStream` | Single-provider (`firstresult`). |
+| `register_execution_result_store` | `ExecutionResultStore` | Single-provider (`firstresult`). |
+| `register_labels_service_or_provider` | `LabelsServiceBase \| LabelsProvider` | Single-provider (`firstresult`). |
+| `register_llm_provider` | `BaseLLMProvider` | Single-provider (`firstresult`). LLM API access for AI-assisted features. |
+
+### LLM provider hook
+
+`register_llm_provider` lets a plugin supply the LLM API client used by AI-assisted
+features (e.g. natural-language query building). The interface lives in
```
haileyok: don't need this "e.g." here.

docs/DEVELOPMENT.md
```diff
@@ -125,6 +125,52 @@ def register_ast_validators() -> None:
...
+
+A direct Anthropic implementation is provided as a reference in
+`example_plugins/src/llm/anthropic_provider.py`, including the request/response and
+`tool_use` translation. It imports the `anthropic` SDK lazily; the SDK is **not**
+declared as a workspace dependency (it conflicts with the pinned
+`typing-extensions`), so install it manually to actually run the provider:
+
+```bash
+uv pip install anthropic
+```
```
haileyok: can we attempt to fix this? the `anthropic` dep should be able to be installed within the `example_plugins` workspace

example_plugins/src/llm/anthropic_provider.py
```diff
@@ -0,0 +1,249 @@
...
+manually to run the provider: ``uv pip install anthropic``.
+
+Configuration (via Osprey ``Config`` or environment):
+
+- API key: ``LLM_ANTHROPIC_API_KEY`` config key, else the ``ANTHROPIC_API_KEY``
+  environment variable (read by the SDK itself if neither is set explicitly).
+- Default model: ``LLM_ANTHROPIC_MODEL`` config key
+  (default: ``claude-3-5-sonnet-latest``).
+- Default max tokens: ``LLM_ANTHROPIC_MAX_TOKENS`` config key (default: ``1024``).
+"""
```
haileyok, line 18: let's use the standard `OSPREY_` prefix that we use elsewhere in this repo

example_plugins/src/llm/anthropic_provider.py
```diff
@@ -0,0 +1,249 @@
...
+    LLMResponse,
+    LLMUsage,
+    ToolCall,
+    ToolDefinition,
+)
+
+if TYPE_CHECKING:
+    from anthropic import Anthropic
+
+DEFAULT_MODEL = 'claude-3-5-sonnet-latest'
```
haileyok: this is a pretty out of date model. let's get the tag for 4.6 sonnet (i believe it is `claude-sonnet-4-6` but please verify)

example_plugins/pyproject.toml
```diff
@@ -7,6 +7,15 @@ dependencies = [
...
 
+# NOTE: The example Anthropic LLM provider (llm/anthropic_provider.py) needs the
+# `anthropic` SDK, but it is intentionally NOT declared here. The workspace pins
+# `typing-extensions==4.6.3`, while every modern `anthropic` requires a newer
+# typing-extensions, so declaring it (even as an optional extra) makes the uv
+# workspace lock unsatisfiable. The provider imports `anthropic` lazily and raises
+# a clear error if it is missing; install it manually to use the provider, e.g.:
+#   uv pip install anthropic
+# (You may also need to relax the typing-extensions pin in the root workspace.)
+
```
haileyok: get rid of this once you fix the llibrary

example_plugins/src/register_plugins.py
```diff
@@ -24,3 +26,14 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
...
+
+@hookimpl_osprey
+def register_llm_provider(config: Config) -> BaseLLMProvider:
+    """Register a direct Anthropic API LLM provider.
+
+    Requires the ``anthropic`` SDK (installed manually, see
+    :mod:`llm.anthropic_provider`) and an API key, but only when the provider is
+    actually invoked.
+    """
+    return AnthropicLLMProvider(config)
```
haileyok: `BaseLLMProvider` shouldn't be a required plugin, and it should be possible for one to not be registered at all. I'd rather us not have the anthropic provider here as a default.

osprey_worker/src/osprey/worker/adaptor/hookspecs/osprey_hooks.py
```diff
@@ -55,6 +56,17 @@ def register_execution_result_store(config: Config) -> ExecutionResultStore:
     raise NotImplementedError('register_execution_result_store must be implemented by the plugin')
 
 
+@hookspec(firstresult=True)
+def register_llm_provider(config: Config) -> BaseLLMProvider:
+    """Register an LLM API provider used by AI-assisted features (e.g. natural-language
+    query building).
+
+    Only the first registered provider is used (``firstresult=True``), mirroring the
+    other single-provider hooks. Return a concrete :class:`BaseLLMProvider`.
```
haileyok: drop the "mirroring other..." language

pyproject.toml
```diff
@@ -282,7 +282,9 @@ module = "pydantic.main"
 implicit_reexport = true
 
 [[tool.mypy.overrides]]
-module = ["jsonpath_rw", "minio.*"]
+# `anthropic` is an optional, lazily-imported SDK used only by the example LLM
+# provider; it is not installed in the workspace (see example_plugins/pyproject.toml).
```
haileyok: rm this comment later

docs/DEVELOPMENT.md
```diff
@@ -125,6 +125,82 @@ def register_ast_validators() -> None:
...
+back on the next `chat()` call.
+
+```python
+from osprey.worker.lib.config import Config
+from osprey.worker.lib.llm.base import BaseLLMProvider
+
+@hookimpl_osprey
+def register_llm_provider(config: Config) -> BaseLLMProvider:
+    return MyLLMProvider(config)
+```
```
haileyok: i think this should actually be its own doc, not part of `DEVELOPMENT.md`

example_plugins/src/register_plugins.py
```diff
@@ -24,3 +24,9 @@ def register_output_sinks(config: Config) -> Sequence[BaseOutputSink]:
 def register_labels_service_or_provider(config: Config) -> LabelsServiceBase:
     """Register a PostgreSQL-backed labels service."""
     return PostgresLabelsService()
+
+
+# NOTE: the `register_llm_provider` hook is intentionally NOT implemented here.
+# An LLM provider is optional, and we don't want one registered by default. See
+# `llm.anthropic_provider.AnthropicLLMProvider` for a reference implementation; a
+# deployment that wants it can add its own `register_llm_provider` hookimpl.
```
haileyok: Clean up this comment. Just note that you can include one here, not the reasoning

haileyok (commented):
Claude:3 **Code review — two-reviewer pass (standard + adversarial), iterated to clean.**

Verdict: **No outstanding bugs.** Both reviewers independently approved the final state; all Medium findings from the first pass are fixed with regression tests.

### Findings & resolutions

| Severity | Where | Finding | Resolution |
| --- | --- | --- | --- |
| Medium | `lib/llm/loop.py` | `run_tool_loop` dispatched tools on the final, limit-tripping iteration and discarded the results — running side effects that can never be sent back to the model | Fixed: break before dispatching on the last allowed iteration, then raise `ToolLoopLimitExceeded`; regression test added |
| Medium | `lib/llm/tools.py` | `dispatch` serialized the handler result *outside* the `try/except`, so a non-serialisable/circular return crashed `run_tool_loop` instead of becoming `is_error` | Fixed: serialize inside the `try`; regression test added |
| Medium | `pyproject.toml` | `typing-extensions` was changed to an unbounded floor (`>=4.14.0`) — a shared, monorepo-wide dep, inconsistent with the exact-pin convention | Pinned exact `==4.15.0` (matches `uv.lock`) |
| Low | `anthropic_provider.py` | Stale comment claimed the 1h cache TTL needs a beta header | Corrected — 1h TTL is GA on the Claude API (set via the `ttl` field, no header) |
| Low | `lib/llm/tools.py` | Handler exception text is echoed to the model (secret-leak vector); `default=None` sentinel undocumented | Documented both |
| Low (accepted) | `anthropic_provider.py` | Benign lazy-client init race under gevent (last-writer-wins) | Accepted as harmless |
| Low (accepted) | `example_plugins/pyproject.toml` | `anthropic>=0.40.0` floor vs exact-pin convention | Left as a floor: example-plugin dep, reproducibility comes from `uv.lock` |

### What's good (both reviewers)
Clean vendor-neutral interface; tests pin real contracts (exact Anthropic request shape, role mapping, full tool_use→tool_result cycle, usage remap, non-dict input degradation, system cache_control block form); correct `firstresult=True`/`None` bootstrap semantics; the tool loop doesn't mutate the caller's messages and terminates correctly; well-scoped dependency/mypy changes.

All local CI gates pass: `ruff`, `ruff format`, `mypy`, full `pre-commit`, `fawltydeps`, and the example-provider unit tests.

<sub>Reviewers ran on the general model (the configured gpt-5.5 reviewer route was failing auth at review time).</sub>

PR #336, opened 2026-06-02
[worker] [coordinator] remove unused `hashring` features

ayubun:
hash ring routing is something that discord uses internally but not something that pertains to any open-sourced osprey code at this time. so, this code can be removed to tidy up the repo a bit

## Summary by CodeRabbit

* **Removals**
  * Removed ring/hash-ring based service discovery, consistent-hash ring, and related rate-limiting and selection APIs.
  * Removed pigeon/gRPC client, interceptors, exceptions, and related utilities and tests.
* **Behavior Changes**
  * Discovery APIs simplified (ring params removed; selector types narrowed).
  * gRPC/worker retry and metric behaviors adjusted (RPC-code annotations removed; fewer cases treated as retryable).

PR #338, opened 2026-06-04
[deps] pin markupsafe<2.1 to unbreak CI

ayubun:
this pr fixes the integration test issues that merged in from #336 . as a follow-up, we should require those to pass for merging

## Summary by CodeRabbit

* **Chores**
  * Pinned `markupsafe` version constraint in dependencies and updated dependency checking configuration.


juanmrad (approved):

haileyok (approved):
thanks!

PR #364, opened 2026-06-16
docker: pin cloud-sdk image

### LINKED ISSUE #363: ci: cloud-sdk/bigtable causing integration tests to fail
The integration test suite started failing consistently around 16:00 UTC on 2026-06-16 across all branches. The failure is always the same:

```
Container bigtable  Error
dependency failed to start: container bigtable is unhealthy
```

It looks like maybe there was a recent cloud-sdk release with an issue: https://github.com/GoogleCloudPlatform/cloud-sdk-docker/pull/651

PR to follow after investigating some more.

### Root cause

The `bigtable` service in `docker-compose.yaml` uses `cloud-sdk:latest`, which is a floating tag. The `cloud-sdk` image does not pre-install `cbtemulator` — it gets downloaded at runtime when `gcloud beta emulators bigtable start` runs.

The healthcheck polls for the `cbtemulator` process:

```yaml
healthcheck:
  test: ["CMD", "bash", "-c", "pgrep -f cbtemulator > /dev/null || exit 1"]
  interval: 10s
  timeout: 5s
  retries: 5
```

With no `start_period`, that gives cbtemulator exactly 50 seconds to download and start before the container is marked unhealthy. When the runner or network is even a little slow, it misses the window.

The `latest` tag also means any upstream image update can silently change startup behavior — which is likely what triggered the regression today.

### Fix

- Pin both `cloud-sdk:latest` references to a specific version + digest (`573.0.0`, the current version)
- Add `start_period: 60s` to the healthcheck so the cbtemulator download time doesn't eat into the retry window

cassidyjames:
The cloud-sdk:latest tag was floating, so any upstream image update could silently change startup behavior. It looks like this may have happened with 573.0.0, so let's pin the previous version until the fix is released.

cbtemulator is also not pre-installed; it gets downloaded at runtime, so the 50s healthcheck window (5 retries × 10s) could be too short when the runner or network was a little slow.

Pin both cloud-sdk references to 572.0.0 with digest, and add start_period: 60s so the cbtemulator download time doesn't count against retries.

Fixes #363

## Summary by CodeRabbit

* **Chores**
  * Updated Docker Compose services to use pinned container image digests instead of the moving `latest` tag.
  * Improved service startup reliability by extending the health check initialization window for the Bigtable service.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -205,7 +205,7 @@ services:
   bigtable:
     hostname: bigtable
     container_name: bigtable
-    image: gcr.io/google.com/cloudsdktool/cloud-sdk:573.0.0@sha256:e032d744e8de1312d7461684ef365777c0036933c39c47c72cfc1adafc22de67
+    image: gcr.io/google.com/cloudsdktool/cloud-sdk:572.0.0@sha256:f67318c2d4719e3d9579ee724dd16b9d028ae916c4919586ae01e9e26d6c1beb
     ports:
       - "127.0.0.1:8361:8361"
     command: >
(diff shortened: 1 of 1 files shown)


julietshen (approved):
LGTM

PR #373, opened 2026-06-17
[diag] Capture native stack of integration hang (do not merge)

haileyok:
Diagnostic only. Runs the suite via the same gevent invocation in the background and an out-of-process watchdog dumps py-spy --native + /proc wchan/stack for all threads if the process is still alive ~180s in, to find the real shutdown blocker. Not for merge.


haileyok:
Diagnostic experiment only — used to capture the native stack of the hang (gevent hub stuck in ev_run/epoll_wait at shutdown). Findings recorded; closing.

PR #377, opened 2026-06-24
fix(ui): tolerate malformed URI escapes in EntityWithPopover

### LINKED ISSUE #376: EntityWithPopover crashes the page when an entity value has malformed URI escapes
## Bug

`osprey_ui/src/components/entities/EntityWithPopover.tsx` calls `decodeURIComponent(entityId)` and `decodeURIComponent(entityType)` unguarded at lines 106–107. `decodeURIComponent` throws `URIError: URI malformed` on any value containing a stray `%` (or any malformed escape), which crashes the React tree and white-screens the whole app.

## Repro

1. `osprey-stress run …` or any setup that produces events with entity values containing `%` not followed by two hex chars.
2. Open the UI and surface those values in the Event Stream — easy to trigger now that #297 falls back to showing every event key.
3. Selecting features in the picker or fast-scrolling the event stream → white screen.

## Stack trace

```
Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at EntityWithPopover (EntityWithPopover.tsx:107)
```

(then the usual React "consider adding an error boundary" message)

## Fix

Tolerate malformed input by wrapping the two calls in a `safeDecode` helper that falls back to the raw value on `URIError`. Ten-line change; PR forthcoming.

## Blocks

#297 (event-stream sensible defaults) shouldn't merge until this is fixed — otherwise the fallback exposes the latent crash to fresh demos.

julietshen:
Closes #376. Unblocks #297.

## Bug

`EntityWithPopover.tsx:106-107` calls `decodeURIComponent(entityType)` and `decodeURIComponent(entityId)` unguarded. `decodeURIComponent` throws `URIError: URI malformed` on any input with a stray `%` or otherwise-malformed escape — which crashes the React tree and white-screens the app.

Latent until now because fresh demo deployments showed "No data for selected features" cards, so most entity values never made it to the popover. PR #297 (event-stream sensible defaults) exposes the crash: scrolling the event stream fast or picking features in the modal triggers it.

## Fix

Wrap the two calls in a `safeDecode` helper that returns the raw value on `URIError`. Ten-line change.

\`\`\`ts
const safeDecode = (s: string): string => {
  try {
    return decodeURIComponent(s);
  } catch {
    return s;
  }
};
\`\`\`

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] \`pnpm run format:check\` clean
- [ ] Live: with #297 also checked out, repro path-3 fast-scroll and path-2 picker selection in EventStream — both previously white-screened, now render the value undecoded (no crash)

🤖 with @claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of entity links with malformed route values, preventing crashes when decoding certain URLs.
  * Popover details and related actions now display using safely decoded entity type and ID values.

### DIFF
--- a/osprey_ui/src/components/entities/EntityWithPopover.tsx
+++ b/osprey_ui/src/components/entities/EntityWithPopover.tsx
@@ -103,8 +103,17 @@ const EntityWithPopover = ({ entityId, featureName, entityType }: EntityWithPopo
   }>();
 
   const isRouteEntity = entityRouteType === entityType && entityRouteId === entityId;
-  const decodedEntityType = decodeURIComponent(entityType);
-  const decodedEntityId = decodeURIComponent(entityId);
+  // `decodeURIComponent` throws on malformed input (stray `%`, etc.). Fall
+  // back to the raw value so a bad ID doesn't crash the whole popover.
+  const safeDecode = (s: string): string => {
+    try {
+      return decodeURIComponent(s);
+    } catch {
+      return s;
+    }
+  };
+  const decodedEntityType = safeDecode(entityType);
+  const decodedEntityId = safeDecode(entityId);
 
   const handleShowLabelDrawer = () => {
     updateShowLabelDrawer(true);


haileyok (approved):
thanks!

Issue #343, opened 2026-06-08
Make Osprey runnable without GCP credentials (tracking)


julietshen:
Osprey has a bunch of features that assume you're running on Google Cloud. If you're not (like adopters trying it out, or people running the demo), those features either silently do nothing or crash.

This issue tracks the work needed to let Osprey run cleanly without Google Cloud credentials.

## Motivation

Surfaced repeatedly while bringing up the JetStream demo (#236) and triaging Dependabot in #219. The pattern is consistent: features ship with one side configured for GCP and the other side missing, or with a GCP client that has no fallback. Specific past discoveries:

- #320 / #286 — `@sentry/browser` initialized only when DSN env was set; DSN never set in default compose; whole error-reporting subsystem silently inert.
- #331 / #332 — `osprey-ui-api` missing the MinIO env vars that `osprey-worker` had; `POST /events/scan` 500'd silently in the default `docker compose up` setup for ~8 months because the broken path wasn't on the demo's golden path.
- Rules visualizer (no issue yet) — `/rules_visualizer/labels_view/` 500'd on `DefaultCredentialsError` because both view handlers eagerly publish telemetry to a GCP Pub/Sub topic that doesn't exist in dev.

### Tier 1 — silently inert features (no error, no UI hint)

- [ ] **Worker analytics publishers** — `osprey_worker/src/osprey/worker/cli/sinks.py:263-264` and `:306-307` register `PubSubPublisher(project='osprey-dev', topic='osprey-analytics')` as output sinks at worker startup. Every rule evaluation pushes an analytics event that goes nowhere in dev.
- [ ] **Webhooks publisher** — `cli/sinks.py:310-311`, `PubSubPublisher(project='osprey-dev', topic='osprey-webhooks')`. The entire downstream webhook subsystem (Slack alerts, custom integrations) is inert without GCP and there's no in-band signal that webhooks aren't firing.
- [ ] **Rules visualizer analytics** — `osprey_worker/src/osprey/worker/ui_api/osprey/views/rules_visualizer.py:45,83` publishes `OspreyRulesVisualizerGenGraphAnalyticsEvent` after every render. Already wrapped best-effort on the demo branch; needs to be upstreamed.

### Tier 2 — half-wired features (loud 500 at first use)

[...]

PR #388, opened 2026-06-29
feat(publisher): degrade PubSubPublisher to noop when GCP credentials are absent

julietshen:
First PR for #343 — making Osprey runnable without GCP credentials. Tracked by #410, and now incorporates the async-worker changes originally split into #411.

## What this changes

Both Pub/Sub publishers now degrade to a noop instead of failing when GCP credentials can't be resolved (the default `docker compose up` experience, or any adopter env without GCP), or when the `DISABLE_GCP_PUBSUB` env var is set. In noop mode no underlying client is built and `publish()`/`stop()` return immediately.

- **Shared credential detection** — `osprey/worker/lib/gcp_credentials.py` exposes `gcp_credentials_available()` (cached + lock-guarded, so `google.auth.default()` is probed at most once per process) and `gcp_pubsub_disabled()` (the `DISABLE_GCP_PUBSUB` env opt-out). Both publishers share this one implementation.
- **Sync `PubSubPublisher`** (`osprey_worker/.../lib/publisher.py`) — auto-detects credentials at construction, honors the env opt-out (short-circuiting before the credential probe), logs a one-time warning, and short-circuits `publish()`/`stop()`.
- **Async `AsyncPubSubPublisher`** (`osprey_async_worker/.../lib/publisher.py`) — the same behavior for the experimental asyncio worker.

## Observability

- **Startup warning** — one parsable log line per publisher when it enters noop mode, naming the inert project and topic.
- **`configuration.errors` metric** — emitted once at startup when credentials are missing (tagged with project, topic, and `reason:gcp_credentials_missing`). The deliberate `DISABLE_GCP_PUBSUB` opt-out stays silent, since it isn't a misconfiguration. There is no per-send metric: the state only changes at startup, and credentials revoked mid-run surface as `publish.failure` spikes rather than a disabled state.

## Not in this PR

Broader GCP-eager modules (`secret_manager.py`, `bigquery.py`, `access_audit_log.py`) remain tracked under #343.

https://claude.ai/code/session_01TEcfpqhza3dPepZUVWw36X


osprey_worker/src/osprey/worker/lib/publisher.py
```diff
@@ -98,8 +125,22 @@ class StrPubSubPublisher(PubSubPublisher):
...
+
+_gcp_credentials_available: Optional[bool] = None
+
+
+def _check_gcp_credentials() -> bool:
+    global _gcp_credentials_available
+    if _gcp_credentials_available is None:
+        try:
+            google.auth.default()
+            _gcp_credentials_available = True
```
chimosky: I'm wondering if a better way to do this would be to check for "Google" in `/sys/class/dmi/id/product_name`, so we can avoid the cost of this network call.

Seeing as we actually aren't using the credentials, only checking for them.

julietshen: oooh good catch. you're right that we only need to check whether creds exist here, not actually use them. I looked into what that network call really costs:

`google.auth.default()` looks for credentials in three places, in order: an environment variable pointing at a key file, then a local gcloud login file, and it only calls to the network if both of those come up empty. So the network call only happens when there are no creds anywhere else, which is the exact case we're trying to catch.

it shouldn't slows down a normal startup that has creds, and I only run the check once per process and remember the answer, so it doesn't happen on every publish.

The reason I'd hold off on reading `/sys/class/dmi/id/product_name`: that tells you whether the machine is a Google Cloud box, which isn't quite the same as "do we have google credentials." Those two usually match, but not always: for example, someone running Osprey outside Google Cloud (say on AWS or their own servers) but still sending to Pub/Sub with a key file has credentials, even though the machine isn't a Google box.

In that case, `google.auth.default()` still finds the key and publishing works. But if we switched to reading the machine type, it would see "not a Google box," assume there are no credentials, and shut publishing off without any error. Since the whole point of #343 is to stop GCP from failing quietly, I'd rather not add another way for that to happen.


If that startup check does turn out to slow down docker compose up, I think the better fix is to just give it a short time limit so it can't get stuck, vs switch to the hardware check and lose the ability to find key-file creds. Happy to do that if you think it's worth it! What do you think?

chimosky: > The reason I'd hold off on reading /sys/class/dmi/id/product_name: that tells you whether the machine is a Google Cloud box, which isn't quite the same as "do we have google credentials." Those two usually match, but not always: for example, someone running Osprey outside Google Cloud (say on AWS or their own servers) but still sending to Pub/Sub with a key file has credentials, even though the machine isn't a Google box.

Yeah, this makes sense.

> In that case, google.auth.default() still finds the key and publishing works. But if we switched to reading the machine type, it would see "not a Google box," assume there are no credentials, and shut publishing off without any error. Since the whole point of https://github.com/roostorg/osprey/issues/343 is to stop GCP from failing quietly, I'd rather not add another way for that to happen.
>
> If that startup check does turn out to slow down docker compose up, I think the better fix is to just give it a short time limit so it can't get stuck, vs switch to the hardware check and lose the ability to find key-file creds. Happy to do that if you think it's worth it! What do you think?

The network cost might be something we need not worry about tbh, putting a limit sounds like a good idea,

osprey_worker/src/osprey/worker/lib/publisher.py
```diff
@@ -72,6 +95,9 @@ def prepare_data(self, data: _PydanticModelT) -> bytes:
         return data.json(exclude_none=True).encode()
 
     def publish(self, data: _PydanticModelT, attributes: dict[str, str] | None = None) -> None:
+        if not self._enabled:
+            metrics.increment(f'{self.__class__.__name__}.publisher.noop', tags=self._tags)
```
ThisIsMissEm: Not sure on the usage of another metric here, and I think the dashboard could probably just surface this as a warning banner or something "Publishing to GCP PubSub is currently disabled due to invalid credentials"

ThisIsMissEm (commented):
The only thing I'm not sure on is the usage of a metric for this state. I think you'd probably want just a config flag fed back into the system. So like allow setup of the publisher to fail, and report it back as a configuration error rather than a metric (one parsable log line vs metric per send), since this is only changing at startup.

It'd also be an idea to add a `DISABLE_GCP_PUBSUB` boolean environment variable that completely opts out of this codepath.

osprey_worker/src/osprey/worker/lib/publisher.py
julietshen: ooooh good idea! I'll work that in

julietshen: i'm actually going to add that separately, since it's a UI change and i want to keep this and #411 focused on the publisher changes

ThisIsMissEm:
@julietshen this gets close to what I was suggesting; here's something you could feed back into claude to adjust the approach, which will lift it up one layer:

### 1. Reuse the existing noop publisher instead of building disable logic into PubSubPublisher

There's already a `NullPublisher` class (`osprey_worker/.../lib/publisher.py:34`) that
does nothing when you publish to it — it exists for exactly this "I don't want to send
anything" case.

Right now the PR teaches `PubSubPublisher` to disable *itself*: it probes for GCP
credentials and flips into a silent mode internally. That gives the transport class a
second job (deciding whether it should run) on top of its real job (sending messages).

The cleaner shape is to make that decision at the point where the publisher is *created*,
and hand back a `NullPublisher` when Pub/Sub is off. Since it's constructed in ~5 places
(`sinks.py:266/307/311`, `data_exporters/validation_result_exporter.py:59`,
`ui_api/osprey/singletons.py:14`), the tidy way is a small factory — e.g.
`make_publisher(project, topic)` — that returns `NullPublisher()` when disabled or when
credentials are absent, and `PubSubPublisher(...)` otherwise. Everything routes through
that one function.

Trade-off worth naming: the current self-disabling approach automatically covers any
*future* call site too, whereas the factory needs new sites to go through it. I think the
factory is still the better call (keeps the sending class simple, reuses code we already
have), but it's a judgment call, not a hard rule.

The credential probe could actually throw and be caught higher up and then the publisher class returned is actually NullPublisher because gcp isn't configured correctly. We're essentially lifting the logic a layer up the stack, rather than focusing on the PubSubPublisher itself.

### 2. Read the flag through the existing config system, not os.environ directly

The codebase has a config helper for reading settings like this:
`config.get_bool('DISABLE_GCP_PUBSUB', False)`. It already understands `"true"`/`"false"`
values. The PR currently reads the raw environment variable by hand
(`os.environ.get('DISABLE_GCP_PUBSUB', '').lower() == 'true'` in `gcp_credentials.py`),
which re-implements what the helper does for free and sidesteps the project's normal way
of handling config. Switching to `get_bool` (from `osprey_worker/src/osprey/worker/lib/config/__init__.py`) keeps it consistent with the rest of the app. There's also `environ_bool` in `osprey_worker/src/osprey/worker/lib/osprey_logging/__init__.py` oddly.

You could potentially even derive this flag from `OSPREY_GCP_PROJECT_ID` being set or not, though there's also `PUBSUB_OSPREY_PROJECT_ID` and `PUBSUB_DATA_PROJECT_ID` so not exactly a consistently named thing.

### 3. Minor: flag naming

`DISABLE_GCP_PUBSUB` reads fine, but note every other flag in the codebase is phrased
positively (`DEBUG`, `TESTING`, `RUN_PROFILER`) — there's no `DISABLE_*` precedent. A
positive form like `PUBSUB_ENABLED` (defaulting to `true`) avoids the "disable = false =
double negative" confusion. Not a blocker either way. There is a `OSPREY_DISABLE_VALIDATION_EXPORTER` but that seems to be the only case of a "disable", I think, not 100% sure.

Also worth keeping in mind: the broader goal (#343) is running Osprey without *any* GCP,
not just Pub/Sub. A Pub/Sub-specific name is fine for this PR, just flagging so we don't
lock ourselves out of a broader `DISABLE_GCP`-style switch later, though I think it may be independently controlled via two PUBSUB_ENABLED and BIGTABLE_ENABLED or something, though both depend on GCP credentials.

ThisIsMissEm:
oh, also, there's only ever GCP PUBSUB, so you could drop the `GCP_` in the env var name.

osprey_worker/src/osprey/worker/lib/utils/gcp_credentials.py
chimosky, line 1: This should be in `osprey_worker/src/ospery/worker/lib/utils`.

julietshen:
Thank you @chimosky and @ThisIsMissEm for the reviews and suggestions! I'm addressing them in my next commit. 

FYI, I'm going to focus on sync since `AsyncPubSubPublisher` has zero production callers right now. I'll follow up on it later

osprey_worker/src/osprey/worker/lib/publisher.py
```diff
@@ -103,3 +114,28 @@ def publish(self, data: str, attributes: dict[str, str] | None = None) -> None:
...
+    misconfiguration, so we log, emit a one-time configuration.errors metric, and
+    fall back to a NullPublisher rather than crash.
+    """
+    # Imported here to avoid pulling the config/singletons stack into this low-level module.
+    from osprey.worker.lib.singletons import CONFIG
+
+    if not CONFIG.instance().get_bool('OSPREY_PUBSUB_ENABLED', False):
+        return NullPublisher()
+    try:
+        return PubSubPublisher(project_id, topic_id)
```
ThisIsMissEm: To retain what was previously here:
```suggestion
        return PubSubPublisher(
          project_id,
          topic_id,
          raise_on_error=raise_on_error,
          max_bytes=max_bytes,
          max_messages=max_messages,
          max_latency=max_latency,
      )
```

Though these perhaps have default values?

osprey_worker/src/osprey/worker/lib/publisher.py
```diff
@@ -103,3 +114,28 @@ def publish(self, data: str, attributes: dict[str, str] | None = None) -> None:
...
+    except DefaultCredentialsError:
+        logger.warning('GCP credentials not detected, publishing disabled (project=%s, topic=%s)', project_id, topic_id)
```
ThisIsMissEm:
```suggestion
        logger.warning('GCP credentials errored, publishing disabled (project=%s, topic=%s)', project_id, topic_id)
```

osprey_worker/src/osprey/worker/lib/publisher.py
```diff
@@ -103,3 +114,28 @@ def publish(self, data: str, attributes: dict[str, str] | None = None) -> None:
...
+    Publishing is opt-in: unless OSPREY_PUBSUB_ENABLED is true, this returns a
+    NullPublisher (the default, no-GCP experience). When it is true but GCP
+    credentials cannot be resolved, PubSubPublisher construction raises; that is a
+    misconfiguration, so we log, emit a one-time configuration.errors metric, and
+    fall back to a NullPublisher rather than crash.
+    """
+    # Imported here to avoid pulling the config/singletons stack into this low-level module.
+    from osprey.worker.lib.singletons import CONFIG
+
+    if not CONFIG.instance().get_bool('OSPREY_PUBSUB_ENABLED', False):
```
ThisIsMissEm: Current behavior would be defaulting to True, I think, if I understand/remember this method correctly:
```suggestion
    if not CONFIG.instance().get_bool('OSPREY_PUBSUB_ENABLED', True):
```

osprey_worker/src/osprey/worker/lib/utils/gcp_credentials.py
julietshen: during the refactor based on feedback from the working group, this is resolved! the shared credential-probe module is gone entirely now. Instead of probing up front, make_publisher just constructs the PubSubPublisher and catches the DefaultCredentialsError, so there's no gcp_credentials module left to relocate.

osprey_worker/src/osprey/worker/lib/publisher.py
julietshen: ty! They do have defaults, but no reason to drop the passthrough. Done in f0b8f6e

osprey_worker/src/osprey/worker/lib/publisher.py
julietshen: agreed: since the construction-catch already returns a NullPublisher when credentials can't be resolved, a no-GCP setup still runs either way. defaulting to True avoids silently disabling publishing for deployments that already have creds. Done in f0b8f6e.

ThisIsMissEm (approved):
`This looks good to me, though I'm wondering if it'd make sense to flip `OSPREY_PUBSUB_ENABLED` to default to `false` and require opt-in? That way we're only producing the metric if something went wrong. 

Though is the `configuration.errors` metric used anywhere else, or is this the only location for it? If so, a `logger.error` might be better than a metric that no one looks at.

julietshen:
thank you!!

`configuration.errors` is new in this PR and this is the only place it's emit, with nothing consuming it yet. I'll drop it and keep the existing `logger.warning`

the reason `OSPREY_PUBSUB_ENABLED` defaults to true is to avoid silently disabling publishing for deployments that already have GCP creds and upgrade without setting the flag. I figured losing outbound data is worse than seeing log errors

ThisIsMissEm:
@julietshen I'd expect people deploying to be reading the release notes prior to deployment, so listing it as a configuration change required before deployment would be okay?

Issue #323, opened 2026-05-30
Remove unused dependencies (likely Discord-era leftovers)


julietshen:
Following the dependency audit that produced #320 (`@sentry/browser` removal), several other packages are pinned in `pyproject.toml` and `osprey_ui/package.json` but never imported anywhere in the source tree. These look like leftovers from when Osprey was a Discord-internal tool and was deployed against Discord's GCP stack.

Cross-checked via grep across `osprey_worker`, `osprey_rpc`, `example_plugins`, and `osprey_ui/src`. Listed in roughly decreasing confidence.

## Python (`pyproject.toml` common deps)

**High confidence — zero imports, safe to remove:**

- [ ] `erlpack` — Erlang term format, fetched from `discord/erlpack` git fork
- [ ] `nostril` — N-gram spam classifier (Discord internal use)
- [ ] `google-api-python-client` — generic GCP, already flagged \"needs investigation\" in fawltydeps ignore_unused
- [ ] `google-auth-httplib2` — same, already flagged
- [ ] `google-cloud-appengine-logging` — Discord App Engine deploy
- [ ] `google-cloud-audit-log` — Discord GCP audit
- [ ] `google-cloud-kms` — Discord uses GCP KMS for key management
- [ ] `google-cloud-logging` — Discord GCP logging
- [ ] `google-cloud-secret-manager` — Discord uses GCP Secret Manager

**Medium confidence — verify before removing:**

- [ ] `traitlets` — IPython transitive that's pinned directly; should be available via `ipython` itself
- [ ] `google-resumable-media` — transitive of `google-cloud-storage` (which IS used); direct pin redundant but harmless
- [ ] `google-auth` — transitive of `google-cloud-bigquery` (which IS used); direct pin redundant but harmless

**DO NOT remove — used dynamically, not via `import`:**

- `blinker` — `sentry-sdk` Flask integration
- `jslog4kube` — gunicorn `--logger-class`
- `grpcio-health-checking`, `grpcio-reflection`, `grpcio-status`, `grpcio-tools` — registered at gRPC server startup

## npm (`osprey_ui/package.json` dependencies)

**High confidence — zero source imports, safe to remove:**

- [ ] `uuid` (9.0.1) — no source imports anywhere; Discord likely used it for request IDs
[...]

PR #391, opened 2026-06-29
build(deps): remove unused Discord-era Python dependencies

### LINKED ISSUE #356: License review: nostril (LGPL 2.1) — possibly unused
`nostril` is listed as a production dependency in the `common` group and is licensed under **LGPL 2.1**. However, it does not appear to be imported anywhere in the codebase — searching for `import nostril` and its primary API (`nonsense`, `is_nonsense`) returns no results.

LGPL is generally permissive enough for use in proprietary software (it allows linking), but the fact that this may be an unused dependency makes removal the simpler path.

**Options:**
- Remove `nostril` from the dependency list if it is genuinely unused
- If it is used (e.g. loaded dynamically or as a transitive requirement), approve LGPL 2.1 and document the rationale

Tracked here to unblock the license-check CI workflow, which currently skips this package pending a decision.

julietshen:
- Fixes #356 

## Summary

Removes the **high-confidence Python cluster** from #323 — nine packages in the common dependency group with zero source imports, all leftovers from the Discord-internal era.

**Direct deps removed (9):**
- `erlpack`, `nostril` (also dropped from `[tool.uv.sources]`)
- `google-api-python-client`, `google-auth-httplib2`
- `google-cloud-appengine-logging`, `google-cloud-audit-log`
- `google-cloud-kms`, `google-cloud-logging`, `google-cloud-secret-manager`

**Dead support code also dropped:**
- `osprey_worker/.../utils/secret_manager.py` — only file importing `google.cloud.secretmanager`, with no callers anywhere in the workspace.
- `googleapiclient.discovery_cache` logger entry in `osprey_logging/logging.yaml` — only existed to silence warnings from the now-removed `google-api-python-client`.

The four corresponding "Needs investigation" entries in `[tool.fawltydeps].ignore_unused` are removed too.

**Impact:** `uv lock` removes 15 packages total (9 direct + 6 transitives: `httplib2`, `humanize`, `plac`, `pyparsing`, `tabulate`, `uritemplate`).

The npm pair (`uuid`, `ajv`) is intentionally left for a follow-up PR, per the issue's suggestion to keep reviewer load small.

## Test plan

- [x] Verified zero imports of each removed package across `osprey_worker`, `osprey_async_worker`, `osprey_rpc`, `example_plugins`
- [x] `uvx fawltydeps` reports no new "unused declared dependency" warnings
- [x] `uv run --frozen ruff check` passes
- [x] `uv.lock` re-resolves cleanly (168 packages, down from ~183)
- [ ] CI test suite passes

Refs #323.

## Summary by CodeRabbit

* **Chores**
  * Cleaned up dependency configuration by removing several unused packages and related dependency-check ignore entries (while keeping required core Google libraries).
  * Removed the secret-management helper and its cloud Secret Manager access support.
  * Simplified logging configuration by dropping the custom log-level override for `googleapiclient.discovery_cache`, allowing the library’s default behavior.

### DIFF
--- a/osprey_worker/src/osprey/worker/lib/osprey_logging/logging.yaml
+++ b/osprey_worker/src/osprey/worker/lib/osprey_logging/logging.yaml
@@ -39,12 +39,6 @@ handlers:
     stream: ext://sys.stderr
 
 loggers:
-  # There is a warning spam that happens that isn't useful from this library
-  # related to file_cache being unavailable.
-  # https://github.com/googleapis/google-api-python-client/issues/299
-  googleapiclient.discovery_cache:
-    level: WARN
-
   pandas_gbq:
     level: WARN
 
--- a/osprey_worker/src/osprey/worker/lib/utils/secret_manager.py
+++ b/osprey_worker/src/osprey/worker/lib/utils/secret_manager.py
@@ -1,8 +1,0 @@
-from google.cloud import secretmanager
-
-
-def fetch_secret(project_id: str, name: str, version: str = 'latest') -> str:
-    gsm_client = secretmanager.SecretManagerServiceClient()
-    secret_reference = gsm_client.secret_version_path(project=project_id, secret=name, secret_version=version)
-    response = gsm_client.access_secret_version(request={'name': secret_reference})
-    return response.payload.data.decode()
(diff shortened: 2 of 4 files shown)


BinaryFiddler (approved):
LGTM!

PR #402, opened 2026-07-02
Add in-app rule authoring backed by a draft rules table

julietshen:
## What

  Adds experimental in-app rule authoring to the UI API. Users draft SML
  rules, the drafts validate against the live engine, and they save to a
  `rule_drafts` Postgres table that the people who operate Osprey can
  reference, edit, and deploy from. No external code host is involved.

  ## Why this changed from the original PR

  The first version of this PR routed drafts through pluggable git
  submission backends (GitHub, local, null) so a "Submit" button opened a
  pull request. Review feedback was to avoid github/gitlab/etc entirely and
  keep drafts inside Osprey. This revision does that: the git backends are
  gone and the draft table is the source of truth.

  ## How it works

  - **Draft table** (`lib/storage/rule_drafts.py`): `RuleDraft` model, one
    row per rule path, upserted on save. Editing a deployed draft moves it
    back to `draft` so the table reflects that the in-flight SML no longer
    matches what was deployed.
  - **Endpoints** (all gated by the `CAN_EDIT_RULE_DRAFTS` ability):
    - `POST /rule-drafts` — re-validate server-side, then upsert.
    - `GET /rule-drafts` — list every draft.
    - `GET /rule-drafts/<id>` — fetch one.
    - `POST /rule-drafts/<id>/deploy` — re-validate, write the SML into the
      configured rules directory (`OSPREY_RULES_LOCAL_PATH`), and mark the
      draft deployed. `wire_into_main: true` also appends a
      `Require(rule=...)` line to `main.sml` so the rule takes effect.
  - The backend-agnostic authoring endpoints are unchanged: `source`,
    `validate`, `vocabulary`, `parse-into-builder`.
  
  ## Deploy model

  Deploy writes into a rules directory the engine's sources provider already
  loads, a filesystem hand-off. A DB-backed `SourcesProvider` that loads
  deployed drafts straight from the table (removing the filesystem
  dependency) is noted in the docs as future direction; this PR keeps the
  filesystem deploy and makes the table the source of truth for drafts.
  
  ## Tests

  `test_rule_drafts.py` covers the authoring endpoints plus the table
  workflow: create / upsert / list / get, and deploy including
  wire-into-main, idempotency, and the 409 / 503 / 404 branches. Docs
  (`docs/user/manage.md`) and the CHANGELOG are updated.

## Summary by CodeRabbit

* **New Features**
  * Added experimental in-app rule authoring with Rule Builder and Code Editor views.
  * Users can create, validate (live engine AST validation), list, and deploy rule drafts.
  * Deploy writes validated rules into the configured local rules directory and can optionally wire drafts into `main.sml`.
  * Added access control gating to edit/deploy rule drafts.
* **Documentation**
  * Expanded user guide with rule draft authoring, API endpoints, and `OSPREY_RULES_LOCAL_PATH` behavior (including 503 when unset).
  * Updated changelog for the experimental feature.
* **Chores**
  * Updated ignore rules for local Docker Compose overrides.


julietshen:
screenshots and more context in https://github.com/roostorg/osprey/discussions/394

julietshen:
Huge thanks to @ThisIsMissEm 

Your feedback made the design a lot simpler. I've reworked the PR to drop the git submission backends  entirely. Draft rules now live in a `rule_drafts` table in Osprey's own Postgres, and there's a deploy endpoint that writes the SML into the configured rules directory (with an option to wire a `Require` line into `main.sml`).

I also think there's  a follow-up to be noted: a DB-backed `SourcesProvider` could load deployed drafts straight from the table and remove the filesystem hand-off altogether, but I kept this PR to the table + deploy so it stays reviewable.

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
```diff
@@ -0,0 +1,110 @@
...
+            draft = session.query(cls).filter(cls.path == path).first()
+            if draft is None:
+                draft = cls(path=path)
+                session.add(draft)
+            draft.rule_name = rule_name
+            draft.sml_source = sml_source
+            draft.summary = summary
+            draft.author_email = author_email
+            draft.status = RuleDraftStatus.DRAFT
+            session.flush()
```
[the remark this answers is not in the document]
ThisIsMissEm: @julietshen you'll probably want to do this. it looks like sqlalchemy supports this: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#insert-on-conflict-upsert

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
```diff
@@ -0,0 +1,124 @@
...
+    One row per rule `path` (upserted on submit), so the table reads as the
+    current set of drafts rather than an append-only history.
+    """
+
+    __tablename__ = 'rule_drafts'
+
+    id: int = Column(BigInteger, primary_key=True, autoincrement=True)
+    path: str = Column(Text, nullable=False, unique=True)
+    rule_name: str = Column(Text, nullable=False)
+    sml_source: str = Column(Text, nullable=False)
```
ThisIsMissEm, line 38: You may also want to add a "cid" column here, or "content id" which the the normalised sml_source hashed, and then embedded that into the rules output written to disk so you can actually compare "rule as deploy" vs "rule as draft" (that way a deployed rule doesn't need to be deleted). Though this probably touches more on having a Rules table with a rule identifier/name, and then multiple Rule Versions which are content hashed, and the rules table points to the currently live version (allowing for time-travel to see how a rule evolved over time)

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
```diff
@@ -0,0 +1,124 @@
...
+    summary: str = Column(Text, nullable=False, default='')
+    author_email: str = Column(Text, nullable=False)
```
ThisIsMissEm, line 42: Might be worth a comment that osprey doesn't actually have a Users table somewhere, instead it's just that you have an email with ACLs applied.

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
```diff
@@ -0,0 +1,124 @@
...
+            session.flush()
+            session.expunge(draft)
+            return draft
+
+    @classmethod
+    def list_all(cls) -> list['RuleDraft']:
+        with scoped_session() as session:
+            drafts = session.query(cls).order_by(cls.updated_at.desc()).all()
+            for draft in drafts:
+                session.expunge(draft)
```
ThisIsMissEm: I think you could probably just do `session.expunge_all()`

ThisIsMissEm: that's based on reading https://docs.sqlalchemy.org/en/21/orm/session_api.html#sqlalchemy.orm.Session.expunge

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+from osprey.worker.ui_api.osprey.lib.abilities import CanEditRuleDrafts, require_ability
+from osprey.worker.ui_api.osprey.lib.auth import get_current_user_email
+
+from ._engine_ast_utils import get_func_identifier
+
+logger = logging.getLogger(__name__)
+
+blueprint = Blueprint('rule_drafts', __name__)
+
+_VALID_PATH = re.compile(r'^[A-Za-z0-9_./-]+\.sml$')
```
ThisIsMissEm: I'd probably remove the `.` from the regexp's range-match, so it's just `A-Za-z0-9_-/` as the range of characters, though the `/` will give you weirdness on windows if someone enjoys deployment pain that much.

ThisIsMissEm: Also validate that the path doesn't begin with `/` and then you can scope it to the configured rules directory. Another way to do this without regex, something like:
```py
from pathlib import Path

base = Path(base_dir).resolve()
target = (base / user_input).resolve()

if not target.is_relative_to(base):
    raise ValueError("escapes base")
```

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+def get_source() -> Any:
+    path = request.args.get('path', '').strip()
+    err = _validate_path(path)
+    if err:
+        return jsonify({'error': err}), 400
+
+    engine = ENGINE.instance()
+    source: Source | None = engine.execution_graph.validated_sources.sources.get_by_path(path)
+    if source is None:
+        return jsonify({'error': f'No source found at {path!r}.'}), 404
```
ThisIsMissEm:
```suggestion
        return jsonify({'error': f'No draft or rule found at {path!r}.'}), 404
```

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+
+    try:
+        sources = Sources.from_dict(spliced)
+    except Exception as exc:
+        # Sources.from_dict asserts on shape (e.g., missing main.sml). Surface as a structured error
+        # so the editor can show "you broke main.sml" without crashing.
+        return jsonify(
+            {
+                'ok': False,
+                'errors': [{'message': str(exc), 'hint': '', 'source_path': path, 'line': 0, 'column': 0}],
```
ThisIsMissEm: Does the exception here from `Source.from_dict` reveal more information?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+        {
+            'features': features,
+            'udfs': udfs,
+            'effects': effects,
+            'source_files': source_files,
+        }
+    )
+
+
+def _revalidate(path: str, source_text: str) -> tuple[Any, int] | None:
```
ThisIsMissEm, line 353: This validation is slightly different to the one at `/rule-drafts/validate`

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+        return jsonify({'error': f'Could not assemble sources: {exc}'}), 400
+    return None
+
+
+@blueprint.route('/rule-drafts', methods=['POST'])
+@require_ability(CanEditRuleDrafts)
+def create_draft() -> Any:
+    """Validate a draft and upsert it into the rule_drafts table (one row per path)."""
+    payload = request.get_json(silent=True) or {}
+    path = (payload.get('path') or '').strip()
```
ThisIsMissEm, line 370: is there a reason not to do:
```suggestion
    path = payload.get('path', '').strip()
```

Same for rule_name below?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+            }
+        ), 409
+
+    draft = RuleDraft.upsert(
+        path=path,
+        rule_name=rule_name,
+        sml_source=source_text,
+        summary=summary,
+        author_email=get_current_user_email(),
+    )
```
ThisIsMissEm, line 413: This should probably happen within a transaction, but may be fine without.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+
+    engine = ENGINE.instance()
+    source: Source | None = engine.execution_graph.validated_sources.sources.get_by_path(path)
+    if source is None:
+        return jsonify({'error': f'No source found at {path!r}.'}), 404
+    return jsonify({'path': source.path, 'contents': source.contents})
+
+
+@blueprint.route('/rule-drafts/validate', methods=['POST'])
+@require_ability(CanEditRuleDrafts)
```
ThisIsMissEm, line 186: You probably want to use `@marshal_with(ClassName)` here like in events.py and the validator in the parent directory: https://github.com/julietshen/osprey/tree/main/osprey_worker/src/osprey/worker/ui_api/osprey/validators

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+
+@blueprint.route('/rule-drafts/vocabulary', methods=['GET'])
+@require_ability(CanEditRuleDrafts)
+def vocabulary() -> Any:
+    engine = ENGINE.instance()
+    sources = list(engine.execution_graph.validated_sources.sources)
+    features = _collect_features(sources)
+    udfs = _collect_udfs()
+    effects = _collect_effects(sources)
+    source_files = sorted(s.path for s in sources)
```
ThisIsMissEm, line 342: You may want to cache these results in memory, and only invalidate that cache on rule publication.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+    try:
+        candidate.relative_to(rules_dir.resolve())
+    except ValueError:
+        return None
+    return candidate
+
+
+def _main_requires(main_sml: str, draft_path: str) -> bool:
+    pattern = re.compile(r"Require\s*\(\s*rule\s*=\s*['\"]" + re.escape(draft_path) + r"['\"]\s*\)", re.MULTILINE)
+    return bool(pattern.search(main_sml))
```
ThisIsMissEm, line 471: Don't you have a Source for the mail.sml file? Then you could modify it via the AST?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm, line 1: I'd suggest separating the AST munging from the validations for requests from the request business logic.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+            suggested.add(path)
+    return sorted(suggested)
+
+
+def _validate_path(path: str) -> str | None:
+    if not _VALID_PATH.match(path):
+        return f'Path {path!r} is not a valid SML source path (must end .sml and contain only [A-Za-z0-9_./-]).'
+    if '..' in path.split('/'):
+        return f'Path {path!r} contains a parent-directory segment.'
+    return None
```
ThisIsMissEm, line 106: this probably isn't the correct logic, since it doesn't constrain to rules directory subpaths.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+    Not,
+    Number,
+    Source,
+    String,
+    UnaryOperation,
+)
+from osprey.engine.ast.grammar import (
+    List as AstList,
+)
+from osprey.engine.ast.sources import Sources
```
ThisIsMissEm, line 27: you could probably introduce a `osprey.engine.ast.builder import Builder` which contains all the builder related logic from this file.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+        author_email=get_current_user_email(),
+    )
+    return jsonify(draft.to_json())
+
+
+@blueprint.route('/rule-drafts', methods=['GET'])
+@require_ability(CanEditRuleDrafts)
+def list_drafts() -> Any:
+    """The draft rules table: every staged draft, newest-edited first."""
+    return jsonify({'drafts': [d.to_json() for d in RuleDraft.list_all()]})
```
ThisIsMissEm, line 421: would it be worth having a `GET /rules` which is all the rules currently persisted to disk?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+            return jsonify({'error': f'wire_into_main requested but main.sml does not exist at {main_path}.'}), 409
+        main_contents = main_path.read_text(encoding='utf-8')
+        if not _main_requires(main_contents, draft.path):
+            main_path.write_text(_append_require(main_contents, draft.path), encoding='utf-8')
+            main_sml_updated = True
+
+    deployed = RuleDraft.mark_deployed(draft_id)
+    result = deployed.to_json() if deployed is not None else draft.to_json()
+    result['main_sml_updated'] = main_sml_updated
+    result['path_on_disk'] = str(target)
```
ThisIsMissEm: this could leak sensitive config to the end-user, potentially consider removing the `rules_dir` prefix from `target`

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,714 @@
...
+        return f'Path {path!r} contains a parent-directory segment.'
+    return None
+
+
+def _current_sources_dict() -> dict[str, str]:
+    engine = ENGINE.instance()
+    return engine.execution_graph.validated_sources.sources.to_dict()
+
+
+@blueprint.route('/rule-drafts/source', methods=['GET'])
```
ThisIsMissEm, line 114: I don't think this is loading a rule draft source specifically, but any deployed rule's source?

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
```diff
@@ -0,0 +1,124 @@
...
+    summary: str = Column(Text, nullable=False, default='')
+    author_email: str = Column(Text, nullable=False)
+    status: RuleDraftStatus = Column(
+        Enum(RuleDraftStatus, native_enum=False, length=32),
+        nullable=False,
+        default=RuleDraftStatus.DRAFT,
+    )
+    created_at: datetime = Column(DateTime(timezone=True), nullable=False, default=_now)
+    updated_at: datetime = Column(DateTime(timezone=True), nullable=False, default=_now, onupdate=_now)
+    deployed_at: datetime | None = Column(DateTime(timezone=True), nullable=True)  # type: ignore[misc]
```
ThisIsMissEm, line 50: What happens if the rule draft is deployed but then deleted from disk by a system admin?

ThisIsMissEm (commented):
Have left a bunch of comments, it's definitely looking better though.

You could probably expose a config endpoint that says whether or not publishing rules is enabled, and then gate that UI feature based on that.

There's also the path sanitisation and validation that should be the same across all codepaths that perform that sort of logic.

osprey_worker/src/osprey/worker/lib/storage/rule_drafts.py
julietshen: oh good catch. Right now the row would still say "deployed" even though the file is gone. I'm treating that as a known gap for now. The content-hash idea you mentioned below is probably the proper fix, since it'd let us compare what's in the table against what's actually on disk

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: Yep, that's on purpose. This endpoint hands back the rule that's already on disk so you can edit an existing rule. A draft's own text comes from the table instead, through GET /rule-drafts/<id>. I'll make the docstring say that clearly.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: i'm changing it to 'no rule found' since this only checks for rules on disk

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: ty for the catch! I pulled both into one shared helper `(_validate_draft_source)`, so /validate and the create/deploy re-check now run the exact same logic and can't drift. The only difference left is what each caller does with the result: /validate returns it as JSON for the editor, and create/deploy turn a failure into a 400.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: yeah, this is on purpose. `payload.get('path', '')` only falls back to '' when the key is missing. If a client sends `{"path": null}` the key exists, so it hands back `None`, and then `.strip()` throws a 500. Wrapping it in `(payload.get('path')` or `'')` turns that None into '' too, so it fails cleanly with a normal 400 instead. Same reasoning for rule_name

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: Agreed, that'd be cleaner and more consistent with events.py and the validators. I'd rather keep this PR focused on getting the draft table working, so I'm going to track moving these endpoints over to @marshal_with schemas as a follow-up rather than do it in this

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
julietshen: The engine does load main.sml as a Source, but deploy writes into the on-disk rules directory `(OSPREY_RULES_LOCAL_PATH)`, which is the deploy target and isn't necessarily the same as what's currently loaded. Right now I read that file and append a Require(...) line as text. You're right that editing it through the AST would be more robust (no formatting surprises, handles weird existing content), so I'll track that as a follow-up and keep the simple append for this first version.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm: That sounds right!

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm: Yeah, probably use the request data model I suggested, will probably clean this up a bit

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm: Yeah, perhaps the URLs should be `/rules/:cid/source` or something for getting an existing on disk rule versus `/rules/drafts/:id/source`

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,747 @@
...
+    errors: list[dict[str, Any]],
+) -> list[str]:
+    """Collect source paths the draft references but doesn't import.
+
+    Pulled from each error's `defined_in_source_paths`. main.sml is the engine
+    entry point and is never importable; the draft can't import itself either.
+    """
+    suggested: set[str] = set()
+    for err in errors:
+        for path in err.get('defined_in_source_paths') or []:
```
chimosky: The pythonic thing here would be to use `[ ]` as a default.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,747 @@
...
+            if path == 'main.sml' or path == draft_path:
+                continue
+            suggested.add(path)
+    return sorted(suggested)
+
+
+def _validate_path(path: str) -> str | None:
+    if not _VALID_PATH.match(path):
+        return f'Path {path!r} is not a valid SML source path (must end .sml and contain only [A-Za-z0-9_/-]).'
```
chimosky: A better error would be `f'Path {path!r} is not a valid SML source path (must end .sml and contain only alphanumeric characters and the special characters "/-").'` might not be this exact term, errors aren't supposed to cryptic.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,747 @@
...
+def _rules_dir_or_error() -> tuple[Path | None, tuple[Any, int] | None]:
+    """The directory deploy writes into (OSPREY_RULES_LOCAL_PATH), or an error tuple.
+
+    Deploying into the engine's own rules source is a filesystem hand-off, the same
+    contract the retired `local` backend used: whatever pipeline already syncs that
+    directory (etcd push, file watcher) activates the rule. A DB-backed
+    SourcesProvider that lets the engine read deployed drafts straight from this
+    table would remove that dependency entirely; see the PR notes.
+    """
+    raw = os.environ.get('OSPREY_RULES_LOCAL_PATH', '').strip()
```
chimosky: For consistency, it'll be great to get this from config.

chimosky (commented):
I'd also like to suggest being able to export/import rules, this way SCM wouldn't be the only way to share rules.

julietshen:
Thank you @ThisIsMissEm and @chimosky for the reviews! I've addressed all your feedback in my latest commits.

chimosky (approved):

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
```diff
@@ -0,0 +1,749 @@
...
+                and get_func_identifier(statement.value) == 'WhenRules'
+            ):
+                call_node = statement.value
+            if call_node is None:
+                continue
+            then_arg = call_node.find_argument('then')
+            if then_arg is None:
+                continue
+            _walk(then_arg.value)
+
```
reitblatt, line 330: This kind of traversal should really be encapsulated with the AST nodes themselves, which is done in ast_utils.iter_nodes/filter_nodes. It would also be much cleaner using them:

```suggestion
    from ast_utils import filter_nodes
    for source in sources:
        # find WhenRules call nodes anywhere in this source (covers both direct Calls and Calls inside Assigns)
        for when_call in filter_nodes(source.ast_root, Call, lambda n: get_func_identifier(n) == "WhenRules"):
            then_arg = when_call.find_argument("then")
            if then_arg is None:
                continue

            # collect all Call nodes inside the `then` expression
            for call in filter_nodes(then_arg.value, Call):
                ident = get_func_identifier(call)
                if ident:
                    seen.add(ident)
```

reitblatt (approved):
Overall I think this looks good enough to unblock iterating on the UI. 

Does Osprey have any existing feature gating tools that we could use to mark this as experimental or unstable, so we can freely change the API as you iterate on the feature?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
reitblatt: Oops I overlooked that this was assigned to @ThisIsMissEm. Any blockers you want fixed before landing?

julietshen:
While we wait for @ThisIsMissEm's response, @reitblatt we do not have feature flags in Osprey - but they are something I do want to explore!

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm: I've covered this in #487 — since I was reorganising all the routes and trying to clean things up.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rule_drafts.py
ThisIsMissEm: Yeah, editing via AST is also not possible right now due to #483, which would mean comments would be removed from the `main.sml`, which isn't desirable.

julietshen:
Thank you for everyone's reviews and work - I'm closing this as this work has been continued on and refactored by @ThisIsMissEm in https://github.com/roostorg/osprey/pull/487 

Let's get this across the finish line 🎉

Issue #410, opened 2026-07-07
Pub/Sub publishers should degrade to noop without GCP credentials (sync + async workers)


julietshen:
Part of #343.

Osprey's Pub/Sub publishers build their client eagerly in `__init__`, so they fail when GCP credentials can't be resolved — breaking the default `docker compose up` experience and any adopter environment without GCP. Both publisher implementations should instead detect missing credentials and degrade to a noop: log a one-time warning, skip building the client, short-circuit `publish()`/`stop()`, and emit a noop metric so the inert state is visible in dashboards.

This issue tracks that work across both worker implementations.

### Sync worker — `PubSubPublisher` (#388)

`osprey_worker/src/osprey/worker/lib/publisher.py`

- [x] Detect missing credentials via `google.auth.default()` (cached, one probe per process)
- [x] Log a one-time warning naming the inert project and topic
- [x] Short-circuit `publish()` and `stop()` in noop mode
- [x] Emit a `PubSubPublisher.publisher.noop` metric, tagged with project and topic
- [x] Unit tests for credential detection and the noop paths

### Async worker — `AsyncPubSubPublisher`

`osprey_async_worker/src/osprey/async_worker/lib/publisher.py` (separate class, tied to the experimental asyncio worker #341)

- [ ] Construction succeeds when GCP credentials are absent (no client built)
- [ ] `publish()` and the flush/`stop()` paths short-circuit cleanly in noop mode
- [ ] One-time warning naming the inert project and topic
- [ ] Noop metric per publish attempt, tagged with project and topic (parity with the sync publisher)
- [ ] Unit tests for credential detection and the noop paths

## Notes

- The credential-detection helper from #388 could be lifted to a shared location both publishers import, rather than duplicated.
- Broader GCP-eager modules (`secret_manager.py`, `bigquery.py`, `access_audit_log.py`) remain tracked under #343 and are out of scope here.

PR #411, opened 2026-07-07
feat(async-publisher): degrade AsyncPubSubPublisher to noop when GCP credentials are absent

julietshen:
Second piece of #410 (part of #343). Stacked on #388 — base this on `julietshen/noop-publisher-fallback`; review after #388 lands.

## What this changes

Applies the noop-when-GCP-credentials-absent behavior from #388 to the async worker's `AsyncPubSubPublisher`, and dedups the credential check so both publishers share one implementation.

- **New `osprey/worker/lib/gcp_credentials.py`** — `gcp_credentials_available()`, the cached, lock-guarded probe of `google.auth.default()` (at most one call per process). Lifted out of the sync `PubSubPublisher` so there is a single source of truth. A future change (e.g. bounding the metadata probe with a timeout) now lands in one place.
- **Sync `publisher.py`** — drops its private `_check_gcp_credentials` copy and imports the shared helper. Credential-check tests move to `test_gcp_credentials.py`; the publisher tests now patch the shared helper directly.
- **Async `AsyncPubSubPublisher`** — detects missing credentials at construction, skips building the `PublisherClient`, logs a one-time warning naming the inert project and topic, and short-circuits `publish_bytes()` (both publish entry points route through it) and `stop()`. Emits `async_pubsub_publisher.publish.noop` per attempt, tagged with project and topic, for parity with the sync side.

## Verification

- `ruff` clean across all changed files; `mypy` clean on both publishers and the shared helper.
- Async publisher tests: 5 passed (2 existing pinned to the enabled path, 3 new: noop construction, publish noop metric, stop short-circuit).
- Sync `lib/tests` are blocked locally by the postgres autouse fixture (same as #388); verified the sync refactor standalone (shared-helper false/cached paths, sync noop publish/stop, enabled path). CI runs the full suite.

## Not in this PR

Broader GCP-eager modules (`secret_manager.py`, `bigquery.py`, `access_audit_log.py`) remain tracked under #343.

https://claude.ai/code/session_01TEcfpqhza3dPepZUVWw36X

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ For more information about each release including git tags and artifacts, see [R
 - Event stream shows sensible defaults so first-load isn't empty ([#297](https://github.com/roostorg/osprey/pull/297) by [@haileyok](https://github.com/haileyok))
 - Replace `react-scripts` with `rsbuild`/`rspack` for UI builds ([#235](https://github.com/roostorg/osprey/pull/235) by [@chimosky](https://github.com/chimosky))
 - Migrate from npm to pnpm via Corepack ([#252](https://github.com/roostorg/osprey/pull/252) by [@haileyok](https://github.com/haileyok))
-- `PubSubPublisher` degrades to a noop when GCP credentials are absent or when `DISABLE_GCP_PUBSUB` is set, so Osprey runs without GCP config instead of failing ([#388](https://github.com/roostorg/osprey/pull/388) by [@julietshen](https://github.com/julietshen))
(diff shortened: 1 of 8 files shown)


ThisIsMissEm:
I'd be inclined to roll this into #388 instead of keeping it as separate. The changes here lift the detection of GCP credentials up a layer, which is probably the right way forwards. You could then have a metric which is "configuration.errors" which would probably more clearly communicate this issue than at publish time, since it's only checked on startup, and credentials being revoked as the worker is running won't drop into a disabled state, they'll just show up as a failure spike

julietshen:
(combining this with #388)

PR #418, opened 2026-07-13
feat(atproto): resolve DID to handle and display name via enrichment UDFs

### LINKED ISSUE #417: Add Jetstream enrichment UDF
**Is your feature request related to a problem? Please describe.**
#236 adds the AT protocol jetstream for load testing and easier demos. However, it doesn't conver did (unique IDs) into usernames or display names. 

**What solution would you like?**
We should also add an enrichment UDF for better demo/testing.

**Describe alternatives you have considered**
We could leave it as just a DID, but then we wouldn't be able to demo searches on usernames or display names (which are common for Osprey)

**Do you have any additional context?**

Image

julietshen:
Closes #417. Stacked on #236 (`hailey/atproto-jetstream-sample`) — review after #236 lands.

## What this adds

JetStream events identify the actor only by DID, so the event stream and rules can't search by handle or display name. This adds two enrichment UDFs in the atproto plugin — `AtprotoHandle(did)` and `AtprotoDisplayName(did)` — that resolve a DID against Bluesky's public, unauthenticated AppView (`app.bsky.actor.getProfile`), over a shared per-DID profile cache, failing soft when the API errors or rate-limits.

**Enrichment is opt-in (off by default).** Each unique DID costs an external API call, which is exactly the dependency you don't want in #236's load-testing scenario. So the default rules run against the raw firehose with no outbound calls; the enrichment lives in `models/enrichment.sml`, which `main.sml` does not import by default. The README documents the one-step opt-in (plus the lexicographically-sorted import) and how to **extend** it with more profile fields (account age, follower/post counts, existing moderation labels) rather than shipping them all wired-in.

It also **removes the dead `IdentityHandle` model** and corrects the README's identity-event shape — JetStream identity events carry only `did`/`seq`/`time`, never a handle (verified against live traffic).

## Verified live

- **Default (off):** rules validate, zero outbound calls, no `Handle`/`DisplayName` in output, `__error_count: 0`.
- **Opt-in (on):** hundreds of handles/display names resolve against the real firehose (e.g. real bsky.social handles), `__error_count: 0`, rate-limited lookups fail soft.
- 31 unit tests pass (fetch, cache, missing-field and transport-error fail-soft).

## Summary by CodeRabbit

* **New Features**
  * Added optional profile enrichment to resolve actor handles and display names from DIDs.
  * Added cached and concurrent profile lookups with one-hour expiry, LRU eviction, and fail-soft handling of unavailable or incomplete profiles.
  * Documented setup, opt-in activation, caching behavior, and rate-limit considerations.

* **Changes**
  * Identity events now expose only DID, sequence, and timestamp fields by default.
  * Removed the identity-handle feature from the default rules configuration.


example_atproto_plugins/src/atproto_plugin/enrichment_udfs.py
```diff
@@ -0,0 +1,99 @@
...
+
+import requests
+from osprey.engine.executor.execution_context import ExecutionContext, ExpectedUdfException
+from osprey.engine.udf.arguments import ArgumentsBase
+from osprey.engine.udf.base import UDFBase
+
+_ATPROTO_CATEGORY = 'ATProto'
+_GET_PROFILE_URL = 'https://public.api.bsky.app/xrpc/app.bsky.actor.getProfile'
+_REQUEST_TIMEOUT_SECONDS = 5
+_CACHE_MAX_SIZE = 10_000
```
haileyok: i think i'd suggest a larger cache size, maybe around 100k?

example_atproto_plugins/src/atproto_plugin/enrichment_udfs.py
```diff
@@ -0,0 +1,99 @@
...
+
+_session = requests.Session()
+# did -> profile dict (the raw getProfile response).
+_profile_cache: 'OrderedDict[str, dict[str, Any]]' = OrderedDict()
```
haileyok: im not super familiar with how to best do this in python and particularly with gevent and locks... i might suggeset to claude to try and use `functools`'s `lru_cache` decorator for this though.

https://docs.python.org/3/library/functools.html#functools.lru_cache

example_atproto_plugins/src/atproto_plugin/enrichment_udfs.py
```diff
@@ -0,0 +1,99 @@
...
+    """Resolves an ATProto DID to its current handle."""
+
+    category = _ATPROTO_CATEGORY
+    execute_async = True
+
+    def execute(self, execution_context: ExecutionContext, arguments: DidArguments) -> str:
+        handle = _profile_or_skip(arguments.did).get('handle')
+        if not handle:
+            raise ExpectedUdfException()
+        return handle
```
haileyok, line 151: hmm. this is interesting and im not exactly sure how to best do this with SML.

it's probably most likely that we're going to want _both_ the handle _and_ the DID in whatever model deals with user events. in that case, especially since these are going to get executed async, i think we'll end up with _two_ requests getting made for the handle. the current locking won't actually stop a second request for being made for a user. this, i think, will double the amount of requests we are making to the api.

there's a couple of things that we could probably do, but the easiest might just be to do something like:

1. Have some UDF called `AtprotoProfile` that returns the profile _object_
2. Use `JsonValue` on the result of `AtprotoProfile` to pull out the handle and display name from the `AtprotoProfile` UDF's result

i think this will work with SML? but im not sure...

example_atproto_plugins/src/atproto_plugin/enrichment_udfs.py
haileyok: tbh we probably also want this to be an expiry cache, since both handles are display names are subject to change

a _preferred_ approach to having the cached items expire though is to have the cache get bust for the given key whenever an identity event (for handles) comes through jetstream _or_ whenever a profile update events comes through. that way we are immediately responding to those things.

this adds some complexity that might not be worth fully ironing out here, but that would be the canonical example of how to do this i think

haileyok:
left some comments. i think this code would work as is, but most of these are suggestions on how we can reduce the number of requests we have to make...

i'd be less hesitant to make sure we get this right if i didn't think that people would actually end up using these in real world use cases, but i figure there's a good chance that they will, so i'd definitely like to at least get right stuff like handling display name/handle updates

julietshen:
Thanks @haileyok, this was really helpful! I ended up keeping the two DID-based UDFs but reworked the cache underneath them to incorporate all your feedback.

- Bumped the size to 100k
- On the double-request thing you flagged: you were right (ofc), and I confirmed it. Since the async UDFs run concurrently in the gevent pool, handle and display name both miss a cold cache and each fire their own getProfile. So I made in-flight fetches shared: if a request for a DID is already running, the second caller just waits on it instead of making its own. This should reduce the calls.

- I also added a 1-hour expiry, since handles and display names change and the old cache held onto them forever. I left a note in the code and README that the proper version of this is to bust a DID's entry when an identity or profile-update event comes through JetStream, but that felt like more than this example needs.

- On `lru_cache`: I tried to use it but it doesn't get us the in-flight dedup or the expiry, so it would've walked back the other two things. The hand-rolled cache is a little more code but does all three in one place.

I seriously considered your AtprotoProfile-returns-the-object idea and i like it but JsonData only reads the raw event data, not another feature's output, so there's no JsonValue-to-pull-from-a-feature to lean on, and a plain dict isn't a valid feature type. To make it work you'd need a custom language type (like we have for `Entity` and `TimeDelta`) plus the two-step SML wiring, which kinda felt heavy for an example. LMK what you think though, i can have Claude do it.

julietshen:
From the discussion in the [August 21 working group](https://github.com/roostorg/osprey/discussions/465), clarified:

- added enrichment.sml
- more information in the README so it's clear that the enrichment UDFs stay registered whenever the plugin is installed (so the compiler can resolve them) but are never invoked until you opt in by importing `models/enrichment.sml` in main.sml

example_atproto_plugins/src/atproto_plugin/enrichment_udfs.py
```diff
@@ -0,0 +1,164 @@
...
+comes through JetStream, which is left out here to keep the example focused.
+
+See the README's "Extending the enrichment" section for how to expose more of the
+profile (account age, follower counts, existing labels) from the same cached fetch.
+"""
+
+import time
+from collections import OrderedDict
+from threading import Event, Lock
+from typing import Any
```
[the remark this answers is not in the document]
julietshen: the source is Bluesky's own public AppView, where handle/displayName are always strings. This plugin exists mainly to run demos and load-test the pipeline against the live firehose, so the input is a first-party, well-formed API rather than untrusted JSON, and per-event validation machinery is overhead the load-testing path specifically doesn't want

PR #420, opened 2026-07-13
ci(deps): make fawltydeps audit member pyprojects + enable undeclared check

julietshen:
Refs #157.

## Problem

fawltydeps was configured with `deps = ["pyproject.toml"]` — the **root** manifest only. Workspace members declare their own runtime deps, so member-declared packages were falsely reported as *undeclared*:

- `flask-cors` — declared in `osprey_worker/pyproject.toml`, imported by `ui_api/osprey/app.py`
- `aiodns` / `pycares` — declared in `osprey_async_worker/pyproject.toml`, imported by the async MX-lookup UDF

They were never actually missing; fawltydeps just wasn't looking at the member manifests.

## Change

- Add the member pyprojects to fawltydeps' `deps`, so it audits the full declared set.
- `ignore_undeclared` the remaining false positives: example_plugins' flat-layout first-party modules (`udfs`, `async_sinks`, `register_async_plugins`) and the typing-only `_typeshed`.
- CI now runs the full `fawltydeps` check (unused **and** undeclared) instead of unused-only, so a future import that isn't declared can't silently depend on a transitive being present.

## Verified

`uv tool run fawltydeps` → "No undeclared or unused dependencies detected" (exit 0), and still exits non-zero on a seeded failure.

## Summary by CodeRabbit

* **Chores**
  * Improved dependency validation to detect both unused and undeclared dependencies.
  * Expanded dependency scanning across workspace package manifests.
  * Added appropriate exclusions for first-party packages and typing support files.


reitblatt (approved):
LGTM.

PR #374, opened 2026-06-17
Replace Highcharts with Apache ECharts in Timeseries chart

reitblatt:
Closes #319.

## Summary

- Removes `highcharts` (proprietary license) and `highcharts-react-official` from `osprey_ui` dependencies.
- Adds `echarts` (Apache 2.0) and `echarts-for-react` as replacements.
- Rewrites `osprey_ui/src/components/timeseries/Timeseries.tsx` to use the ECharts API while preserving all existing behaviour: column chart, purple bars (`#8e5ea2`), datetime x-axis with per-granularity label formatting, integer y-axis, drag-to-select range (brush) that calls `updateExecutedQuery`, and the "No data available" empty-state overlay.

## License

`echarts` is published under the **Apache 2.0** license, resolving the proprietary-dependency concern raised in #319.

## Security

`pnpm audit` shows no CVEs introduced by this change. The existing findings in the audit output all pre-date this PR and are in unrelated transitive dependencies (`rc-picker > moment`, `eslint > minimatch`, `axios > form-data`, etc.).

## Behaviour notes

- **Timezone**: ECharts' `type: 'time'` axis renders timestamps in the browser's local timezone by default, matching the previous `useUTC: false` + `timezone: dayjs.tz.guess()` Highcharts setup.
- **Drag-to-select**: The chart activates ECharts' `lineX` brush mode on mount so the user can drag directly to select a date range, mirroring Highcharts' `zoomType: 'x'`. The cursor changes to a crosshair on hover, making the interaction discoverable.
- **No watermark**: ECharts does not embed a credits badge, so no `credits: { enabled: false }` equivalent is needed.

## Test plan

- [X] `pnpm typecheck` passes (verified locally)
- [X] `pnpm format:check` passes (verified locally)
- [ ] CI `ui-quality` job passes (typecheck + build on Node 22)
- [X] Chart renders with purple bars and correct axis labels: Screenshot 2026-06-17 at 15 24 46

- [X] Granularity selector re-fetches data: 
Screenshot 2026-06-17 at 15 50 12

- [X] Drag on chart updates the query date range: 
Screenshot 2026-06-17 at 15 51 27

- [X] Empty state ("No data available") overlay appears when there is no data: 
Screenshot 2026-06-17 at 15 52 00

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Updates**
  * Timeseries chart date-range selection now uses brush-based interaction instead of zoom
  * Date formatting adjusted in Timeseries component

PR #441, opened 2026-07-22
Bump form-data to 3.0.5 (CR/LF escaping fix)

julietshen:
Bumps the vulnerable `form-data@3.0.4` (reached transitively via `jsdom`, our Jest test DOM) to `3.0.5`, which escapes CR/LF/quotes in field names — the security fix that the form-data half of #361 was carrying.

Because pnpm re-resolved the tree, this also picks up sibling patch bumps within existing ranges: `ws` 7.5.10 → 7.5.13 (closes a ws advisory too), `nwsapi`/`acorn` (jsdom), and `form-data` 4.0.5 → 4.0.6 + `hasown` on axios's runtime line. All patch-level.

Supersedes #361, whose highcharts 8.1.0 → 9.0.0 bump is now moot since #374 replaced Highcharts with Apache ECharts.

### DIFF
--- a/osprey_ui/pnpm-lock.yaml
+++ b/osprey_ui/pnpm-lock.yaml
@@ -884,6 +884,11 @@ packages:
     engines: {node: '>=0.4.0'}
     hasBin: true
 
+  acorn@8.17.0:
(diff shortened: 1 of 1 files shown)


juanmrad (approved):
:shipit:

PR #455, opened 2026-08-14
 Fix excessive CPU load in generate_test_data.sh

reitblatt:
## Summary
- `generate_test_data.sh` was spawning a brand-new `kafka-console-producer` JVM process per message (once/sec), paying full JVM startup + classpath scan + producer init/metadata fetch every iteration, forever — this pegged the `osprey-kafka-test-data-producer` container at 180-200% CPU cycling every 1-2s under `docker top`.
- Now starts a single long-lived `kafka-console-producer` process and streams generated messages to it over a FIFO, instead of forking a new producer per line.
- Same JSON generation logic (`generate_action`, `template.json` substitution) and ~1 msg/sec cadence are preserved.
- `cleanup()` now also closes the FIFO write end and kills the persistent producer on SIGINT/SIGTERM, so it isn't left orphaned.

## Test plan
- [x] Restarted the running `osprey-kafka-test-data-producer` container with the fix applied (volume-mounted `example_data`).
- [x] `docker top` shows a single persistent `ConsoleProducer` JVM process across multiple message sends, instead of a new one per second.
- [x] `docker stats` shows CPU settling to ~2-3% instead of 180-200%.
- [x] Sent SIGTERM to the script and confirmed the trap fires ("Stopping data generation...", exit 0) and the producer JVM is killed rather than orphaned.
- [x] confirmed test data generation still works: 
Screenshot 2026-08-14 at 15 50 16

## Checklist

- [ ] Tests pass locally
- [ ] `uv run ruff check .` passes (no unused imports or other lint errors)
- [ ] `uv tool run fawltydeps --check-unused --pyenv .venv` passes (no unused dependencies)
- [X] Updated `CHANGELOG.md` with my changes, if notable (refer to [Keep a Changelog](https://keepachangelog.com/en/2.0.0/) conventions)

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved test data generation by streaming actions through a persistent messaging process.
  * Reduced process startup overhead when generating large volumes of test data.

* **Reliability**
  * Added orderly cleanup of temporary resources and background processes.
  * Improved recovery and restart behavior for the test data producer service.

* **Security**
  * Improved command handling to reduce risks associated with special characters and unexpected input.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -222,6 +222,7 @@ services:
     entrypoint:
       - /bin/bash
     command: ["/osprey/example_data/generate_test_data.sh"]
+    restart: unless-stopped
 
   postgres:
     hostname: postgres
--- a/example_data/generate_test_data.sh
+++ b/example_data/generate_test_data.sh
@@ -56,10 +56,21 @@ build_kafka_command() {
     echo "$cmd"
 }
 
+# FIFO used to stream messages into a single long-lived producer process
+FIFO="$(mktemp -u /tmp/kafka_producer_fifo.XXXXXX)"
+producer_pid=""
+
 # Function to handle cleanup on script termination
 cleanup() {
     echo
     echo "Stopping data generation..."
+    # Close our write end of the FIFO so the producer sees EOF and exits
+    exec 3>&- 2>/dev/null
+    if [ -n "$producer_pid" ] && kill -0 "$producer_pid" 2>/dev/null; then
+        kill "$producer_pid" 2>/dev/null
+        wait "$producer_pid" 2>/dev/null
+    fi
+    rm -f "$FIFO"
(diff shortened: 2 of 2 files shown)


julietshen:
Thank you for this! While reviewing I realized that the old way (which was excessive) also had a self-healing property. Are we okay if this container has no restart policy? Should we add a restart to the osprey-kafka-test-data-producer service?

reitblatt:
> Thank you for this! While reviewing I realized that the old way (which was excessive) also had a self-healing property. Are we okay if this container has no restart policy? Should we add a restart to the osprey-kafka-test-data-producer service?

Makes sense. Added a restart policy (unless-stopped).

julietshen (approved):

PR #468, opened 2026-08-25
[gevent worker] prevent late monkey-patching in pytest

ayubun:
## 🤖 summary
prevents gevent test modules from monkey-patching an already-running plain pytest process, while preserving the gevent worker's pre-patched integration runner and forced-exit workaround

## related issues/tasks
- https://github.com/roostorg/osprey/pull/415
- https://github.com/roostorg/osprey/pull/467
- https://github.com/roostorg/osprey/pull/469

## changes made (˶ᵔ ᵕ ᵔ˶)
- avoid calling `gevent.monkey.patch_all()` after plain pytest has started
- tie the forced-exit workaround to pytest starting under the pre-patched gevent runner
- add subprocess regression coverage for every import-time monkey-patch site

## models used
- anthropic/claude-opus-4-8: ~50%
- openai/gpt-5.6-sol: ~30%
- anthropic/claude-haiku-4-5: ~10%
- anthropic/claude-fable-5: ~10%

## testing
- plain collection completed with 1,286 worker tests
- the pre-patched runner still collected all 5 discovery tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test startup handling to prevent failures caused by late concurrency setup.
  * Prevented unnecessary process termination when tests modify concurrency behavior during execution.
  * Improved reliability when collecting tests with standard pytest.

* **Tests**
  * Added regression coverage confirming representative test modules collect successfully.
  * Verified collection completes without lock-release errors or collection failures.
  * Confirmed normal collection summaries are produced consistently.

### DIFF
--- a/osprey_worker/tests/test_gevent_collection.py
+++ b/osprey_worker/tests/test_gevent_collection.py
@@ -1,0 +1,65 @@
+"""Regression tests for plain pytest collection of gevent-patched test modules.
+
+Some test modules under ``osprey_worker`` used to call ``monkey.patch_all`` at
+import time. Under plain ``pytest`` (not the pre-patched integration runner)
+that late patch mutates global gevent state after threading/ssl are already
+imported, raising ``RuntimeError: cannot release un-acquired lock`` during
+collection. The ``pytest_sessionfinish`` force-exit hook then saw socket as
+patched and called ``os._exit``, hiding the terminal traceback and leaving a
+bare exit code 2.
+
+These tests run plain ``pytest --collect-only`` in a child process (so gevent
+is NOT pre-patched) and assert collection succeeds with a normal summary.
+"""
+
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+# repo root is three parents up from this file
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+
+# one representative target per import-time gevent patch site:
+# - discovery/tests/test_discovery.py patches in the module body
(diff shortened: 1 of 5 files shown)


osprey_worker/tests/test_gevent_collection.py
```diff
@@ -0,0 +1,63 @@
...
+    combined = result.stdout + result.stderr
+    assert result.returncode == 0, (
+        f'plain `pytest --collect-only` on {target} exited {result.returncode}\n'
+        f'stdout:\n{result.stdout}\nstderr:\n{result.stderr}'
+    )
+    # the specific late-monkey-patch failure and pytest's collection-error
+    # markers must be absent, and a healthy run reports the collected count
+    assert 'cannot release un-acquired lock' not in combined, combined
+    assert 'error during collection' not in combined, combined
+    assert 'collected in' in result.stdout, result.stdout
```
julietshen: i think this will also say this if it's empty, no?

julietshen (approved):
LGTM, tiny nit on a false green if there are zero tests

osprey_worker/tests/test_gevent_collection.py
ayubun: 🤖 good catch ! the exit-code assert above does block the zero-test case (pytest exits 5 with `no tests collected in ...`), but the summary assert shouldn't rely on a different assert to be meaningful. tightened it to `re.search(r'\d+ tests? collected in', ...)` so an empty collection can never match, ty <3

PR #469, opened 2026-08-25
[ci] isolate compose test stack

ayubun:
## 🤖 summary
isolates the compose test dependency graph from host ports and fixed container names so test stacks can coexist across worktrees

## related issues/tasks
- https://github.com/roostorg/osprey/pull/415
- https://github.com/roostorg/osprey/pull/467
- https://github.com/roostorg/osprey/pull/468

## changes made (˶ᵔ ᵕ ᵔ˶)
- reset inherited container names for test dependencies
- remove host port publications that the test runner does not use
- preserve internal service names, ports, healthchecks, and dependency ordering

## models used
- anthropic/claude-opus-4-8: ~45%
- anthropic/claude-fable-5: ~25%
- openai/gpt-5.6-sol: ~20%
- anthropic/claude-haiku-4-5: ~10%

## testing
- ran the full compose integration suite alongside an existing Osprey stack
- 1,308 tests completed with 0 errors and 0 failures, then all run-owned resources were removed

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test environment reliability by allowing multiple test stacks to run concurrently without container name or host port conflicts.
  - Updated service networking so test components communicate through the compose project network.

### DIFF
--- a/docker-compose.test.yaml
+++ b/docker-compose.test.yaml
@@ -2,18 +2,29 @@
 # Usage: docker compose -f docker-compose.yaml -f docker-compose.test.yaml --profile test <command>
 
 services:
+  # reset fixed names and host ports so test stacks can coexist
+  # dependencies communicate over the compose project network
   postgres:
+    container_name: !reset null
     ports: !reset []
     volumes: !reset []
 
+  osprey-kafka:
+    container_name: !reset null
+    ports: !reset []
+
   minio:
+    container_name: !reset null
+    ports: !reset []
     volumes: !reset []
 
+  snowflake-id-worker:
+    container_name: !reset null
+    ports: !reset []
+
   etcd:
-    container_name: etcd
+    container_name: !reset null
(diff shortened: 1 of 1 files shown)


docker-compose.test.yaml
```diff
@@ -2,18 +2,29 @@
...
     ports: !reset []
     volumes: !reset []
 
+  osprey-kafka:
+    container_name: !reset null
+    ports: !reset []
+
   minio:
+    container_name: !reset null
+    ports: !reset []
```
julietshen, line 18: not a blocker, but flagging for @cassidyjames  I think this is the first time we use the !reset tag, which requires Docker Compose ≥ 2.24.4 and the docs don't pin a version, so we should mention a minimum version in the docs somewhere

julietshen (approved):
LGTM

docker-compose.test.yaml
cassidyjames: Thanks! Addressed in #471

PR #473, opened 2026-08-26
fix: overly-broad regular expressions

cassidyjames:
## Description

I noticed these when looking into something else, and they seemed easy enough to fix! For context, `[A-z]` matches punctuation between `Z` and `a`, making it overly broad and usually not what was intended. Instead, we should explicitly match `[A-Za-z]`.

## Checklist

- [x] Tests pass locally
- [ ] Updated `CHANGELOG.md` with my changes, if notable (refer to [Keep a Changelog](https://keepachangelog.com/en/2.0.0/) conventions)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text interpolation handling for word characters, including letters, numbers, and underscores.
  * Corrected leetspeak matching to exclude punctuation characters that were previously treated as letters.

### DIFF
--- a/osprey_ui/src/components/entities/RichDescription.tsx
+++ b/osprey_ui/src/components/entities/RichDescription.tsx
@@ -7,8 +7,8 @@ interface RichDescriptionProps {
   features: { [featureName: string]: string };
 }
 
-const INTERPOLATE_REG_EX = /{([A-Za-z0-9_]+)}/;
-const SPLIT_REG_EX = /(:?{[A-Za-z0-9_]+})/;
+const INTERPOLATE_REG_EX = /{(\w+)}/;
+const SPLIT_REG_EX = /(:?{\w+})/;
 
 const RichDescription = ({ description, features }: RichDescriptionProps) => {
   const interpolatedDescription = description.split(SPLIT_REG_EX).map((part, i) => {
--- a/osprey_ui/src/components/entities/RichDescription.tsx
+++ b/osprey_ui/src/components/entities/RichDescription.tsx
@@ -7,8 +7,8 @@ interface RichDescriptionProps {
   features: { [featureName: string]: string };
 }
 
-const INTERPOLATE_REG_EX = /{([A-z0-9_]+)}/;
-const SPLIT_REG_EX = /(:?{[A-z0-9_]+})/;
+const INTERPOLATE_REG_EX = /{([A-Za-z0-9_]+)}/;
+const SPLIT_REG_EX = /(:?{[A-Za-z0-9_]+})/;
 
 const RichDescription = ({ description, features }: RichDescriptionProps) => {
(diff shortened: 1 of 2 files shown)


osprey_worker/src/osprey/engine/stdlib/udfs/string.py
```diff
@@ -235,7 +235,7 @@ class StringCleaningArguments(StringArguments):
 )
 
 # sub for l33t -> leet
-_L33T_THREES_SUB_PATTERN: re.Pattern[str] = re.compile(r'([A-z]?)(3+)([A-z]?)', flags=re.IGNORECASE)
+_L33T_THREES_SUB_PATTERN: re.Pattern[str] = re.compile(r'([A-Za-z]?)(3+)([A-Za-z]?)', flags=re.IGNORECASE)
```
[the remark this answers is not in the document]
cassidyjames: I think that's fine; the point of the pattern is to match cases like `l33t` = `leet`, so `Pİ3` becoming `Pİe` is expected; the regex is doing its job of replacing `3` used in place of `E` around other letters.

osprey_ui/src/components/entities/RichDescription.tsx
```diff
@@ -7,8 +7,8 @@ interface RichDescriptionProps {
   features: { [featureName: string]: string };
 }
 
-const INTERPOLATE_REG_EX = /{([A-z0-9_]+)}/;
-const SPLIT_REG_EX = /(:?{[A-z0-9_]+})/;
+const INTERPOLATE_REG_EX = /{([A-Za-z0-9_]+)}/;
+const SPLIT_REG_EX = /(:?{[A-Za-z0-9_]+})/;
```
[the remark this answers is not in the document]
cassidyjames: A test here seems overkill to me but I'll let others chime in.

reitblatt: Even simpler

```suggestion
const INTERPOLATE_REG_EX = /{(\w+)}/;
const SPLIT_REG_EX = /(:?{\w+})/;
```

osprey_worker/src/osprey/engine/stdlib/udfs/string.py
reitblatt, line 239: With IGNORECASE [A-Za-z] is equivalent to [a-z]. I can't decide which one is clearer/simpler, so do what you think best. But I think it's worth at least adding a comment to explain that the unicode matching behavior is expected.

reitblatt (approved):
LGTM, minor nitpicks aside.

osprey_worker/src/osprey/engine/stdlib/udfs/string.py
cassidyjames: Ack! Added a comment.

PR #481, opened 2026-09-01
fix: use a pinned Git source for jsonpath-rw

### LINKED ISSUE #480: jsonpath-rw dep uses URL-based path
Pointed out by @ThisIsMissEm:

> small thing I noticed, we currently use a url based installation for jsonpath-rw:
> ```
> jsonpath-rw = { url = "https://github.com/kennknowles/python-jsonpath-rw/archive/6f5647bb3ad2395c20f0191fef07a1df51c9fed8.tar.gz" }
> ```
> when we could be doing what our other git dependencies are doing:
> ```
> jsonpath-rw = { git = "https://github.com/kennknowles/python-jsonpath-rw.git", rev="6f5647bb3ad2395c20f0191fef07a1df51c9fed8" }
> ```

mah1104ahm:
## Summary

- replace the URL-based `jsonpath-rw` source with a pinned Git source
- preserve the exact upstream commit revision
- refresh only the affected lockfile entry

## Verification

- `uv lock --check`
- `uv tree --package jsonpath-rw`
- `git diff --check`

Closes #480

## Summary by CodeRabbit

* **Chores**
  * Updated the dependency source configuration to use a Git repository at the same pinned revision.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -58,7 +58,7 @@ common = [
     { name = "grpcio-tools", marker = "platform_machine == 'x86_64'", specifier = "==1.49.1" },
     { name = "gunicorn", git = "https://github.com/discord/gunicorn.git?rev=979efdcb918daa536d8923668241c6e6bf1edb58" },
     { name = "intervals", specifier = "==0.9.2" },
-    { name = "jsonpath-rw", url = "https://github.com/kennknowles/python-jsonpath-rw/archive/6f5647bb3ad2395c20f0191fef07a1df51c9fed8.tar.gz" },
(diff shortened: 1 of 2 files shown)


uv.lock
```diff
@@ -1182,20 +1182,12 @@ wheels = [
...
     { name = "six" },
 ]
-sdist = { hash = "sha256:df5e3ee97cf486d9872b301982218ab27976647d4c2ffabe459f8b3f8d2c65c8" }
-
-[package.metadata]
-requires-dist = [
-    { name = "decorator" },
-    { name = "ply" },
-    { name = "six" },
-]
```
ThisIsMissEm, line 1198: I'm not sure on these changes? They look possibly unrelated.

cassidyjames: uv records this metadata for URL sources, but not git source since the commit hash is built in. I confirmed the same results locally.

cassidyjames (approved):

PR #486, opened 2026-09-01
chore: Add changelog guidance to AGENTS.md, symlink CLAUDE.md

cassidyjames:
I've noticed some persistent issues with agents trying to follow the Keep A Changelog format, so hopefully this helps.

As long as I was here, I corrected some other minor outdated info:
- Highcharts was replaced with ECharts
- We don't include `v` in release tags

Also, apparently Claude doesn't always check AGENTS.md, so symlink CLAUDE.md to it. 🙃

## Summary by CodeRabbit

* **Documentation**
  * Updated UI charting guidance to reference ECharts.
  * Added changelog conventions, including formatting and release-entry guidance.
  * Clarified release tag formatting to use semantic versions without a `v` prefix.


ThisIsMissEm (commented):
LGTM!

PR #403, opened 2026-07-02
Add rule authoring/editing UI backed by the rule-drafts API

julietshen:
Stacked on #402. Review that first; this PR's diff will show #402's commits until it merges.

Adds a RuleEditorPage at `/rules/new` and `/rules/edit?path=<file>`, built on the rule-drafts API.

- **Rule Builder view**: form-based conditions and outcomes compiled to SML client-side, with the vocabulary (features, UDFs, effects) fetched from the running engine. Editing an existing file first round-trips it through `/rule-drafts/parse-into-builder`; files outside the builder's subset fall back to the Code Editor.
- **Code Editor view**: plain SML editing for anything the builder can't represent.
- Live validation against the engine's AST validator (600ms debounce) with structured inline errors and a missing-imports quick fix.
- Submit posts to `/rule-drafts/submit` and surfaces the review URL returned by the configured backend.
- RulesPage gains an "Add rule" button, per-rule Edit links, and a pending-drafts banner fed by `/rule-drafts/pending` (soft-fails when no backend is configured).

Security/correctness hardening in the SML generator: backslashes and newlines are escaped (not just quotes) before interpolation, the rule name is constrained to an SML identifier before it reaches the generated file and the submit gate, and switching between Builder and Code Editor now carries content across so a tab flip can't submit stale or empty source.

## Checklist

- [x] `pnpm run typecheck`, `pnpm run lint` (0 errors), `pnpm run format:check` pass
- [ ] Updated `CHANGELOG.md`

## Summary by CodeRabbit

* **New Features**
  * Added an experimental rule authoring workflow with visual Rule Builder and code editing modes.
  * Added live validation, error and warning feedback, missing-import suggestions, and draft save/edit support.
  * Added draft listing and editing actions from the Rules Registry.
  * Added secure draft deployment, including optional wiring into the main rules file.
* **Documentation**
  * Documented rule authoring, validation, permissions, and deployment configuration.
  * Added the feature to the unreleased changelog.
* **Chores**
  * Excluded local Docker Compose override files from version control.

Issue #58, opened 2025-11-18
Code-free rules management through UI


cassidyjames:
**Is your feature request related to a problem? Please describe.**

If an org wants to deploy new rules, currently they have to deploy them as code.

**What solution would you like?**

An admin at an org using Osprey should be able to add a rule through the Osprey UI (thus storing the rules in a database versus code). 

**Do you have any additional context?**

This idea has been floating around for a while, I figure I can at least get it into the issue tracker for @vinaysrao1. This will almost certainly need to be broken down into smaller tasks. :)

cassidyjames:
@julietshen to spec this out a bit more and work on requirements

cassidyjames:
A few updates here:
- #394 has some proof-of-concept design work from @julietshen
- #487 from @ThisIsMissEm refactors @julietshen's rule drafts API/back-end work towards this
- #490 from @ThisIsMissEm refactors @julietshen's rule drafts front-end work towards this

It might make sense to break this issue down into a few sub-issues or file new ones for the next steps that we see coming out of the above work?

ThisIsMissEm:
@cassidyjames the main thing I'm looking for feedback on is if it's working as expected for people, and if the UI changes make sense.

ThisIsMissEm:
@julietshen I think we need to make a decision here: do we support writing rules to disk at all, or do we only support database backed rules and etcd backed rules?

The writing rules to disk currently means they aren't actually deployed until the worker and ui-api are reloaded, since the StaticSourcesProvider does not reload rules ever. It loads them once on application boot with: `StaticSourcesProvider(sources=Sources.from_path(Path(rules_path_str)))`

The original concept I think does call for rules entirely stored in the database and not written to disk, which would also probably mean we need a way to import rules from disk into the database (we already have this for etcd). Then we'd have three rules backends:
- Static — doesn't support rule editing at all
- Database — supports rule editing, but requires a new provider
- Etcd — supports rule editing kinda — the drafts are in the database, the deployments target "files" in etcd (which isn't currently implemented in #487)

Given where we are at right now with those two pull requests (#487 and #490) how do we move this from "needs design" to "ready to implement"?

ThisIsMissEm:
I do also now have a pretty solid design for a fully database-backed SML files setup, but I'd like to get some decisions on the above before I review it for correctness and share.

PR #487, opened 2026-09-02
Refactor rule drafts APIs

ThisIsMissEm:
- Related to #58

## Description

This pull request supersedes #402, taking the original work, and refactoring it heavily to reorganise code to hopefully improve maintainability, whilst also tackling a few bugs found and improving user experience.

I've added in:

- `GET /config` exposes whether rule deployment is available at all, and whether the current user can deploy.
- Changed the permissions so `GET /rules` and `GET /rules/source` both need a new `CAN_VIEW_RULES` permission, instead of reusing `CAN_VIEW_DOCS`
- Moved the draft APIs to `/rules/drafts/` from `/rule_drafts/`
- Refactored to make better use of AST utilities when working with the drafts.
- Refactored to organise code better (splitting out validators and schemas for requests and responses, moving common code into `lib/`, and moving inline code in views into respective locations)
- Fixed an issue with the query tests randomly failing (different tests in the code would create residue in the database, that meant they weren't correctly isolated)
- Fixed an issue in osprey validators where we couldn't reuse them in rule authoring
- Added a `CAN_DEPLOY_RULES` permission, since it's foreseeable that operators may want to split permissions between who can author rules and who can deploy those.
- Expanded test coverage quite a bit
- **Config change:** deploy now writes to `OSPREY_RULES_PATH` (the directory the engine already loads from) instead of a separate `OSPREY_RULES_LOCAL_PATH`. Nothing set the old key, so deploy previously returned 503 in every deployment. Deploy is unsupported under the etcd sources provider; use `osprey push-rules`.
- Deploy failures now return meaningful statuses (503 unconfigured, 409 missing or unparseable `main.sml`, 400 path escape) instead of unhandled 500s.
- `wire_into_main` parses `main.sml` rather than regex-matching it: a commented-out `Require` no longer counts as live, and an unparseable `main.sml` refuses the deploy rather than being appended to.
- `rule_name` is now unique in the database; a collision returns 409.
- Status codes are chosen so a client can tell which body shape it holds: 400 malformed request, 422 uncompilable SML, 409 conflict, 404 missing, 503 misconfigured deployment.
- **Breaking for #403:** draft `id` is now a string, and deploy returns `{rule, main_sml_updated, path_on_disk}` rather than the row itself.
- `OspreyEngine` exposes `udf_registry` and `validator_registry` properties.
- Rewrote the Rule Authoring section of `docs/user/manage.md`.

## Checklist

- [x] Tests pass locally
- [x] Updated `CHANGELOG.md` with my changes, if notable (refer to [Keep a Changelog](https://keepachangelog.com/en/2.0.0/) conventions)

## Summary by CodeRabbit

* **New Features**
  * Added experimental rule authoring with draft creation, validation, visual builder support, storage, retrieval, and deployment.
  * Added a rules catalog, source viewing, builder vocabulary, and optional wiring into `main.sml`.
  * Added separate permissions for viewing, editing, and deploying rules.
  * Osprey can start without cloud credentials; publishing is opt-in and gracefully degrades when unavailable.

* **Bug Fixes**
  * Unsupported attribute-based calls now return validation errors instead of server errors.
  * Invalid rule paths are rejected with clearer validation feedback.

* **Documentation**
  * Documented rule authoring, permissions, validation, and deployment workflows.


osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Every refusal deploy_rule can raise carries the status it should map to -- an
+    # unconfigured rules directory (503), a main.sml that's missing or doesn't parse
+    # (409), a path that escapes the rules directory (400). Without this they'd all
+    # surface as an unhandled 500.
+    try:
+        deployment = deploy_rule(draft, wire_into_main=request_model.wire_into_main)
+    except DeployError as exc:
+        logger.warning('deploy of draft %s failed: %s', request_model.draft_id, exc)
+        return jsonify({'error': str(exc)}), exc.status
```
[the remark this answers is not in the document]
ThisIsMissEm: I don't think DeployError exposes any information, can you recheck?

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
```diff
@@ -0,0 +1,231 @@
...
+            }
+        conditions.append(cond)
+    if not conditions:
+        # Builder needs at least one row to render anything sensible; matching the EMPTY_BUILDER_MODEL default.
+        conditions = [{'feature': '', 'operator': '==', 'rhs': '', 'rhs_is_feature': False}]
+
+    outcomes: list[dict[str, Any]] = []
+    if when_rules_call is not None:
+        rules_any_arg = when_rules_call.find_argument('rules_any')
+        if rules_any_arg is not None and isinstance(rules_any_arg.value, AstList):
```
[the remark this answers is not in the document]
ThisIsMissEm: Moved to issue: https://github.com/roostorg/osprey/issues/488

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
```diff
@@ -0,0 +1,231 @@
...
+
+The builder's expressible subset is deliberately narrow, so most of this module
+is about detecting the shapes it *cannot* represent and saying why. When a file
+falls outside the subset the user is sent to the Code Editor instead — a wrong
+round-trip would silently rewrite a rule, which is worse than refusing.
+"""
+
+from __future__ import annotations
+
+from typing import Any
```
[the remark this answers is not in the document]
ThisIsMissEm: Moved to issue: https://github.com/roostorg/osprey/issues/489

.vscode/extensions.json
ThisIsMissEm, line 1: We already included `.idea` but I was having issues with vscode using jedi instead of pylance, where autocomplete and other code intellisense was breaking, this at least fixes it somewhat. (I do have a config file that would fix more fully)

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
ThisIsMissEm, line 1: This code should be essentially as it was in `views/rules.py` in #402, I've just moved it to a dedicated file

CHANGELOG.md
```diff
@@ -8,8 +8,13 @@ For more information about each release including git tags and artifacts, see [R
 
 ## [Unreleased]
 
+### Added
+
+- Experimental in-app rule authoring: draft SML rules validate against the live engine, save to a `rules` table, and deploy into the configured rules directory. This does not currently support etcd. ([#402](https://github.com/roostorg/osprey/pull/402) by [@julietshen](https://github.com/julietshen), [@thisismissem](https://github.com/thisismissem))
```
ThisIsMissEm, line 13: @julietshen this is the line that was causing some strife earlier, in #402, once rebased against main, it ended up under 1.1.0 which was unreleased at the time, but git's merging does some weird things sometimes where it doesn't quite pick up on author's intent.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,275 @@
...
+from osprey.engine.ast.grammar import Source
+from osprey.worker.lib.storage.rules import Rule, RuleNameTaken
+from osprey.worker.ui_api.osprey.lib.abilities import CanDeployRules, CanEditRules, require_ability
+from osprey.worker.ui_api.osprey.lib.auth import get_current_user_email
+from osprey.worker.ui_api.osprey.lib.marshal import marshal_with
+from osprey.worker.ui_api.osprey.lib.rule_builder import parse_into_builder_model
+from osprey.worker.ui_api.osprey.lib.rule_deployment import MAIN_SML_PATH, DeployError, deploy_rule, plan_deployment
+from osprey.worker.ui_api.osprey.lib.rule_validation import validate_draft_source
+from osprey.worker.ui_api.osprey.schemas.rule_validation import DraftValidation, ValidationMessage
+from osprey.worker.ui_api.osprey.schemas.rules import DraftList, DraftSummary, RuleRecord
```
ThisIsMissEm, line 15: In case you're wondering why we have `DraftSummary` here, it's to reduce overfetching the drafts continuously from the client side, as previously it included all the other fields on a draft, which isn't necessary for just a list of drafts. (list vs get one tends to be minimal vs full)

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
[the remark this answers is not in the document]
ThisIsMissEm: Yeah, fixing this would be a much larger refactor I think, hence allowing it this time but marking it as a follow up.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,275 @@
...
+        return jsonify({'error': f'No draft with id {request_model.draft_id}.'}), HTTPStatus.NOT_FOUND
+
+    # Same statuses as deploy for the conditions that make planning impossible: an
+    # unconfigured rules directory (503), a path escaping it (400). A plan that
+    # succeeded where the deploy 503s would be the mismatch this exists to prevent.
+    try:
+        plan = plan_deployment(draft)
+    except DeployError as exc:
+        logger.warning('deploy plan for draft %s failed: %s', request_model.draft_id, exc)
+        return jsonify({'error': str(exc)}), exc.status
```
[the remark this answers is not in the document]
ThisIsMissEm: Would this be a reasonable fix too:

```
 class DeployError(Exception):
          status = 500
          @property
          def message(self) -> str:
              return str(self)
```
and then using `exc.message` here instead? The errors are meant to be user-facing in that file.

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_deployment.py
```diff
@@ -0,0 +1,355 @@
...
+    parser discards comments entirely, so they aren't in the tree to print. Appending
+    to the raw text leaves everything above the new line byte-for-byte untouched.
+    """
+    suffix = f"\nRequire(rule='{rule_path}')\n"
+    if main_sml and not main_sml.endswith('\n'):
+        suffix = '\n' + suffix
+    return main_sml + suffix
+
+
+def deploy_rule(rule: Rule, *, wire_into_main: bool = False) -> RuleDeployment:
```
julietshen, line 229: Using Claude to help review since this is a chonky PR!

a "successful" deploy never reaches the running engine; it takes effect only on process restart. Deploy is allowed only when `OSPREY_RULES_PATH` is set, which is exactly the config where the engine runs `StaticSourcesProvider`, whose `set_sources_watcher` is a no-op (`sources_provider_base.py:41-48`);
 `_handle_updated_sources` never fires. The one provider that hot-reloads (etcd) is explicitly refused. So deploy writes the file, appends the `Require`, marks the row `DEPLOYED`, returns 200, and traffic keeps evaluating the old ruleset with nothing signaling a restart is required.

This would essentially not work, I think

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Rule names are global in SML, and validation above only sees *deployed* rules, so
+    # it cannot know another draft already claimed this one. `rule_name` is unique, so
+    # let the insert be the check: a pre-flight SELECT would cost a query on every
+    # create to answer a question the write answers for free -- and wouldn't be
+    # authoritative anyway, since two concurrent creates can both find nothing.
+    try:
+        draft = Rule.upsert(
+            path=request_model.path,
+            rule_name=request_model.rule_name,
```
julietshen, line 124: (Claude assisted)

i think the unique rule_name guards client metadata, not the rule the SML actually defines

`rule_name` is free-text (identifier regex only), never compared to the X = Rule(...) in the source (the repo's own test creates `rule_name='FirstRule'` over SML defining `SomeRule`). Two drafts can each define `Shared = Rule(...)` under different `rule_names`, both pass the unique check and both validate (a draft's SML isn't visible to another's validation), then deploying both writes two files defining Shared → duplicate-rule engine-load failure.

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rules.py
```diff
@@ -0,0 +1,240 @@
...
+            ):
+                call_node = statement.value
+            if call_node is not None:
+                when_rules_total += 1
+                rules_any_arg = call_node.find_argument('rules_any')
+                if rules_any_arg is not None and isinstance(rules_any_arg.value, AstList):
+                    for item in rules_any_arg.value.items:
+                        if isinstance(item, Name):
+                            whenrules_ref_count[item.identifier] = whenrules_ref_count.get(item.identifier, 0) + 1
+                continue
```
julietshen, line 190: looks like draft edits are last-write-wins with no compare-and-set. `upsert` uses `ON CONFLICT (path) DO UPDATE` with no `WHERE`/version guard, so two concurrent editors of the same path silently collide with each other which will lose an update, and `author_email` flips to whoever saved last, with no error.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/catalog.py
```diff
@@ -0,0 +1,39 @@
...
+from osprey.worker.ui_api.osprey.lib.rules import get_rule_source, list_rules
+from osprey.worker.ui_api.osprey.validators.rules import RuleSourceQuery
+
+from . import blueprint
+
+logger = logging.getLogger(__name__)
+
+
+@blueprint.route('/rules', methods=['GET'])
+@require_ability(CanViewRules)
```
julietshen, line 16: re-gating `GET /rules` to `CAN_VIEW_RULES` breaks the Rules page for existing adopters. The frontend loads `/rules` and renders the nav unconditionally (no ability gating), and only `super_user.json` / `rule_author.json` grant the new permission. Any ACL that had `CAN_VIEW_DOCS` but not `CAN_VIEW_RULES` would now gets 401 until each ACL is hand-edited

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+@blueprint.route('/rules/drafts', methods=['POST'])
+@require_ability(CanEditRules)
+@marshal_with(CreateDraftRequest)
+def create_draft(request_model: CreateDraftRequest) -> Any:
+    # main.sml is the engine entry point; a draft never replaces it wholesale.
+    # Reported as a DraftValidation rather than an ad-hoc {'error': ...} so the editor
+    # has one failure shape to render for this endpoint.
+    if request_model.path == MAIN_SML_PATH:
```
julietshen: From Claude:

The deploy path assumes filename capitalization is treated the same way in the database and on disk, but that isn't always true, which lets deploys overwrite the wrong file.

 Two cases:
  
The `main.sml` guard only blocks the exact lowercase spelling. In `views/rules/drafts.py:93` the reserved-name check is `request_model.path == MAIN_SML_PATH` where `MAIN_SML_PATH = 'main.sml'`. A draft named `Main.sml` passes the guard. On a case-insensitive filesystem (the macOS default) that then writes to the same file as `main.sml`, overwriting the engine's entry point, and the next engine load breaks.

Two rules whose paths differ only in case collide. The DB stores `path` as case-sensitive unique (storage/rules.py), so `spam.sml` and `Spam.sml` are two separate draft rows. But on a case-insensitive filesystem they resolve to one file, so deploying both silently overwrites one with the other. `_resolve_within uses resolve()`, which does not normalize case, so nothing catches this before the write.

This only triggers on case-insensitive filesystems (macOS dev machines; typically not Linux prod), but anyone developing on a Mac will hit it, and the narrow guard is worth tightening regardless.

Suggested fixes: compare the reserved-name guard case-insensitively (e.g. `path.lower() == 'main.sml'`), and decide on one rule for path casing, either reject non-lowercase paths in `VALID_PATH` or treat path as case-insensitive in the DB and on disk so the two agree.

julietshen (changes requested):

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
julietshen: Same root cause (VALID_PATH is too permissive) also bites paths with redundant slashes: rules//spam.sml is accepted, the file is written to the normalized rules/spam.sml, but _append_require writes Require(rule='rules//spam.sml') verbatim and _main_requires matches that verbatim string. So the Require names a path the file isn't at, and a sibling draft at the normalized path appends a duplicate Require. Tightening VALID_PATH (single slashes, lowercase) fixes both this and the casing issue above.

ThisIsMissEm:
Okay, have split out a bunch of commits from this pull request into separate PRs (they're still in #487 as it needs all those pieces to be worked on easily):
- https://github.com/roostorg/osprey/pull/495
- https://github.com/roostorg/osprey/pull/496
- https://github.com/roostorg/osprey/pull/497
- https://github.com/roostorg/osprey/pull/499
- https://github.com/roostorg/osprey/pull/500

Also opened: 
- https://github.com/roostorg/osprey/issues/498

osprey_worker/src/osprey/worker/lib/tests/test_utils.py
```diff
@@ -20,6 +21,32 @@
...
+    and the compose files pointed the tests and the dev stack at the same `osprey`
+    database. Running the suite therefore deleted the local development database, and
+    the failure surfaced later and elsewhere -- as `osprey-ui-api` refusing to start --
+    rather than as a test failure.
+    """
+    if not created_by_this_run:
+        return False
+
+    database_name = urlsplit(url).path.lstrip('/')
+    return database_name.endswith(_TEST_DATABASE_SUFFIX)
```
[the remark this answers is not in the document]
ThisIsMissEm: I don't think this can be tested given it's test setup code?

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
```diff
@@ -0,0 +1,241 @@
...
+
+def _condition_from_value(node: Any, feature: str, operator: str) -> dict[str, Any] | None:
+    """Build a builder Condition row from an RHS node, returning None if the node
+    isn't a literal or a bare Name (the only RHS shapes the builder supports)."""
+    if isinstance(node, Name):
+        return {'feature': feature, 'operator': operator, 'rhs': node.identifier, 'rhs_is_feature': True}
+    if isinstance(node, String):
+        return {'feature': feature, 'operator': operator, 'rhs': node.value, 'rhs_is_feature': False}
+    if isinstance(node, Number):
+        return {'feature': feature, 'operator': operator, 'rhs': str(node.value), 'rhs_is_feature': False}
```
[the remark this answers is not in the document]
ThisIsMissEm: `rule_builder.py` is mostly moved code from #402, and I've not modified it. I'd probably be inclined to say these are follow-up issues?

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_deployment.py
```diff
@@ -0,0 +1,355 @@
...
+        main_contents = main_path.read_text(encoding='utf-8')
+        if not _main_requires(main_contents, rule.path):
+            pending_main_sml = _append_require(main_contents, rule.path)
+
+    target.parent.mkdir(parents=True, exist_ok=True)
+    target.write_text(rule.sml_source, encoding='utf-8')
+
+    main_sml_updated = pending_main_sml is not None
+    if pending_main_sml is not None:
+        main_path.write_text(pending_main_sml, encoding='utf-8')
```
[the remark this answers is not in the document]
ThisIsMissEm: I'm not entirely sure on this, it sounds correct potentially, but I'm not sure what the production outcomes are here? @julietshen I might need someone on the discord side to advise on the right approach here.

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_builder.py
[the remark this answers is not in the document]
ThisIsMissEm: Yes please

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
```diff
@@ -0,0 +1,275 @@
...
+        return jsonify(result.dict()), HTTPStatus.BAD_REQUEST
+
+    return jsonify(result.dict())
+
+
+@blueprint.route('/rules/drafts/<int:draft_id>', methods=['GET'])
+@require_ability(CanEditRules)
+@marshal_with(GetDraftRequest)
+def get_draft(request_model: GetDraftRequest) -> Any:
+    draft = Rule.get_one_with_id(request_model.draft_id)
```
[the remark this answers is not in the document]
ThisIsMissEm: I believe it's intentional, such that someone working on a draft can share it with a coworker and get help on the SML?

osprey_worker/src/osprey/worker/ui_api/osprey/validators/rules.py
```diff
@@ -0,0 +1,169 @@
...
+    A non-object body is discarded rather than merged. `get_json` happily returns a
+    list, string or number, and spreading one of those into a dict is either a
+    `TypeError` or nonsense.
+    """
+
+    @classmethod
+    def marshal(cls: Type[T], flask_request: Request) -> T:
+        body = flask_request.get_json(silent=True)
+        if not isinstance(body, dict):
+            body = {}
```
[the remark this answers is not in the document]
ThisIsMissEm: This one is ending up being a bit more of a broader fix overall, as we've multiple different places where we have marshalling code for json request bodies, and that's probably not correct.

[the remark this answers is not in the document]
ThisIsMissEm: Have opened: #502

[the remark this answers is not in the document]
ThisIsMissEm: Yeah, the idea is that this PR would be rebased after #502 lands.

run-tests-local.sh
```diff
@@ -0,0 +1,35 @@
...
+# no reason to rely on that.
+#
+# Usage: ./run-tests-local.sh [pytest args...]
+#   ./run-tests-local.sh osprey_worker/src/osprey/worker/ui_api -q
+
+set -euo pipefail
+
+export PYTHONPATH=.
+export TESTING=true
+export POSTGRES_HOSTS='{"osprey_db":"postgresql://osprey:FoolishPassword@127.0.0.1:5432/osprey_test"}'
```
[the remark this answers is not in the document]
ThisIsMissEm: This is a file for local development, and the secrets already exist in cleartext in the docker compose files.

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_deployment.py
ThisIsMissEm: @julietshen this matches the behaviour you originally had, however, I do agree that it's probably a problem. I'm just not sure how to fix it correctly.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
ThisIsMissEm: I think it'd fail because before deploy there's a check that ensures everything compiles together. So you'd deploy one and then go to deploy the other and encounter an error that would need to be fixed.

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rules.py
ThisIsMissEm: This is true, but I'd consider it a follow up to wire through CID on save, such that we can detect if the current version is different from the CID the edit is being made on.

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/catalog.py
ThisIsMissEm: Yeah, that's why I listed it as a breaking change. I think it's better to make this one time fix, since `require_ability` can't do an OR between two abilities. The rules definitely aren't documents which are a separate endpoint.

ThisIsMissEm: We could arguably expand the `GET /config` endpoint to include a full permissions object, which would allow us to hide menu items you don't have access to?

osprey_worker/src/osprey/worker/ui_api/osprey/views/rules/drafts.py
ThisIsMissEm: This is fair, but only on OS X which isn't what the code typically runs under, though it would catch anyone who happens to deploy docker containers on a case-sensitive filesystem — debian usually defaults to case-insensitive filesystems though, afaik.

I don't disagree with the `path.lower()` here, just that it's not as big an issue as claude might make it out to be.

ThisIsMissEm: Have fixed this now, both at a database level and at a request level

osprey_worker/src/osprey/worker/ui_api/osprey/lib/rule_deployment.py
ThisIsMissEm: @julietshen I think fixing this is a much harder challenge than it may appear:
- We don't currently implement deploying to etcd (a write operation)
- We also are needing to reload across different containers, so just writing to disk like we do now wouldn't be easily picked up by another container on a different compute node without network attached storage.

We *could* arguably use LISTEN/NOTIFY in postgresql to let the ui-api notify the worker that the rules have changed, and then the worker would do it's own write to disk and reload the rules for the engine?

osprey_worker/src/osprey/worker/lib/tests/test_utils.py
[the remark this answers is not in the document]
ThisIsMissEm: Okay, this is being taken care of in #499

osprey_worker/src/osprey/worker/lib/storage/rules.py
```diff
@@ -0,0 +1,282 @@
...
+    # Requiring paths to be lowercase is what makes the two agree, and it is stated here
+    # rather than only in `validators.rules.VALID_PATH` because the API is not the only
+    # way to write a row and the invariant belongs where the data does. Uniqueness then
+    # needs nothing special: if every path is lowercase, `path`'s existing unique
+    # constraint is already case-insensitive uniqueness.
+    #
+    # A `CheckConstraint` rather than lowercasing on write, because silently storing
+    # something other than what the caller asked for is how a row and a filename come to
+    # disagree in the first place.
+    __table_args__ = (CheckConstraint('path = lower(path)', name='ck_rules_path_is_lowercase'),)
```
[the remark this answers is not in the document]
ThisIsMissEm: I don't think a database can do such a check constrain without a complex function?

osprey_worker/src/osprey/worker/lib/tests/test_utils.py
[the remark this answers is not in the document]
ThisIsMissEm: Have updated #499

Issue #489, opened 2026-09-02
Type the Rule Builder's AST walk instead of using Any


ThisIsMissEm:
> [!Note]
> This is introduced in #402 / #487, but the refactor might end up being a bit too large to ship all at once, so I'm inclined to do it as a follow up task.

## Problem

`lib/rule_builder.py` annotates most of its AST-walking helpers with `Any`:

```python
def _condition_from_value(node: Any, feature: str, operator: str) -> dict[str, Any] | None
def _parse_condition(node: Any) -> dict[str, Any] | None
def _parse_outcome_arg(arg: Any) -> dict[str, Any] | None
def _parse_outcome(node: Any) -> dict[str, Any] | None
def _parse_into_builder_dict(source: Source) -> dict[str, Any]
```

The module is entirely about deciding which AST shapes the Rule Builder can represent,
so `Any` removes type checking from exactly the code whose job is shape discrimination.
mypy cannot tell that `node.left` is only valid after the `isinstance(node,
BinaryComparison)` narrowing, nor catch an attribute that no grammar node has.

## Suggested change

Annotate against the grammar rather than `Any`:

- Node parameters take the shared AST base (`osprey.engine.ast.grammar` exposes the
  node hierarchy), or a narrow union of the shapes each helper actually accepts —
  `_parse_condition` handles `UnaryOperation | Call | BinaryComparison`.
- `_parse_outcome_arg` takes `Keyword` rather than `Any`; it only ever reads `.name`
  and `.value`.
- The `dict[str, Any]` return values could become the `schemas/rule_builder.py` models
  directly (`BuilderCondition`, `BuilderOutcome`), which would delete the intermediate
  dict layer that `BuilderParseResult.parse_obj` currently re-validates.

That last point is the larger win and probably the right shape: the module builds plain
dicts, then hands them to pydantic to be validated into models that already exist.
Returning the models directly would make the walk type-checked end to end.

## Notes

- Raised by CodeRabbit on #487, against the project guideline "Avoid new `# type:
  ignore` and `Any` in Python".
- Mostly pre-existing: the helpers moved into `lib/rule_builder.py` from
  `views/rule_drafts.py:551-611` in #487 with their annotations unchanged.
  `_parse_into_builder_dict` was `_parse_into_builder_model` there; the rest kept their
  names.
- Not urgent — this is type coverage, not a defect. But `#TODO` worth doing alongside
  the `WhenRules` validation fix, since both touch the same functions and the stricter
  types would have made that bug visible.

PR #490, opened 2026-09-02
Rule Authoring UI

ThisIsMissEm:
- Related to #58

## Description

This is the UI side of the work in #489, it was originally the #403 branch, but I've ended up heavily reworking the UI to hopefully feel better.

@julietshen if you're wanting to check out a branch to take for a spin, then it's this branch. You can add a `osprey_worker/src/osprey/worker/lib/acls/dev_acl_assignments.json` to give you two users to test with:
```
{
  "local-dev@localhost": { "roles": ["SUPER_USER"] },
  "author@example.com": { "roles": ["RULE_AUTHOR"] }
}
```

## Checklist

- [ ] Tests pass locally
- [ ] `uv run ruff check .` passes (no unused imports or other lint errors)
- [ ] `uv tool run fawltydeps --check-unused --pyenv .venv` passes (no unused dependencies)
- [ ] Updated `CHANGELOG.md` with my changes, if notable (refer to [Keep a Changelog](https://keepachangelog.com/en/2.0.0/) conventions)

### DIFF
--- a/osprey_ui/src/App.tsx
+++ b/osprey_ui/src/App.tsx
@@ -6,6 +6,7 @@ import { getApplicationConfig } from './actions/ConfigActions';
 import UdfDocsView from './components/docs/UdfDocsView';
 import BulkJobHistoryView from './components/bulk_job_history/BulkJobHistory';
 import { FeaturesPage } from './components/features/FeaturesPage';
+import { RuleEditorPage } from './components/rules/RuleEditorPage';
 import { RulesPage } from './components/rules/RulesPage';
 import RulesVisualizerView from './components/rules_visualizer/RulesVisualizer';
 import EntityViewBar from './components/entities/EntityViewBar';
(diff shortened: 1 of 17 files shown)


chimosky:
Tested 0df941f, some comments;

When using `LabelAdd` as an outcome, there's no help for the arguments to the label, `delay_action_by` and `expires_after` throw this error when the wrong value is used;
```typescript
argument `delay_action_by` to `LabelAdd` has incompatible type
has type `int`, expected `TimeDeltaT | None`
rules/more_rule.sml:18:58
```

while `apply_if` throws this for the wrong value;
```typescript
argument `apply_if` to `LabelAdd` has incompatible type
has type `int`, expected `RuleT | None`
rules/more_rule.sml:18:51
```

The UI doesn't offer a way to create these either, it would be great if these were handled, could be as more options to fill for each argument, or a helper description that shows how to fill in these values which is then used to create the right argument in the API.

When adding a label as an effect, the `label` argument should be a dropdown that shows the current labels, this makes it easier for the rule author as they won't have to know every label that exists, and the variables defined here don't apply to labels 

`__missing_effect__` shouldn't throw a validation error as it's used as a placeholder, it can be confusing for a user as that's the first error they see when they try creating a Rule, before performing any action.
The error is also easier for an engineer to understand, and a Rule author won't always be an engineer, so it's not intuitive. 
A better error might be "No effect for {Rule}".

After saving a draft - also happens with deployed rules-, the edit page for the rule becomes strictly SML, and that defeats the goal of the Rule Authoring UI.

PR #502, opened 2026-09-04
Simplify request body handling

ThisIsMissEm:
Read request bodies in one place

## What this changes

Every marshaller in `ui_api` decided for itself what a request body was. There were four
implementations of that decision and they disagreed with each other:

| | `JsonBodyMarshaller` | `ViewArgAndJsonBodyMarshaller` | `EntityMarshaller` |
|---|---|---|---|
| body under a non-JSON content type | parsed (`force=True`) | parsed | **dropped in silence** |
| body that is a JSON list | parsed, then `parse_obj` errors | **`TypeError` → 500** | **`TypeError` → 500** |
| no body at all | 400 | 400 | defaults applied |

This adds one `json_object_body(request)` that answers the question once, and gives
`JsonBodyMarshaller` an `overrides()` hook so a marshaller says only *where its other
fields come from*. A marshaller written against the base never touches the body API, so
it cannot get this wrong.

## Bugs fixed

- **A reachable 500.** `EntityMarshaller` did `{**body, **entity}`, so any client could
  crash an entity endpoint by POSTing `[1,2,3]`. Now a 400.
- **Silently dropped bodies.** `get_json()` returns `None` for a body sent without
  `Content-Type: application/json`, which read as "no body" — so a request's contents
  were discarded and the endpoint reported success for work it had not done. Now a 400.
- **Dead code with a live bug.** `ViewArgAndJsonBodyMarshaller` had no users and merged
  the *body* last, so a body key could override the URL's own path parameter — a request
  to `/thing/1` carrying `{"thing_id": 2}` would have acted on 2 while the URL, the logs
  and any audit record said 1. Deleted, and replaced by
  `ViewArgAndOptionalJsonBodyMarshaller`, which merges view args last so the URL wins.
  That replacement has no callers on `main` yet; the rules-API branch is its first
  consumer, and it ships here so the broken version can be removed rather than left in
  place for something to find.

## Behaviour changes

`JsonBodyMarshaller` no longer passes `force=True`. A body sent to any endpoint using it
(create draft, validate, bulk actions, druid queries) must now declare
`Content-Type: application/json`. Requests that already send the header are unaffected;
every existing test did.

An empty body is no longer a parse failure. `json.loads('')` raises like malformed JSON
does, but a caller that set the content type and sent nothing is a caller who sent
nothing — whether that was allowed is then the model's business, via its own required
fields.

## Testing

`lib/tests/test_marshal.py` covers the matrix directly against the helper rather than
through whichever endpoint happens to use it: 8 shapes that read as an object or as
nothing, 8 that are refused, plus the `overrides` hook and `marshal_with`'s 400. The
tests use a bare Flask app — no database, engine, or authenticated user.

---

## The underlying problem this does not fix

**A request is being treated as one flat namespace.** Body, view args and query args are
merged into a single dict before validation, so a model cannot say which field comes from
where. Everything awkward here follows from that:

- The source of a field is encoded in the *class* (which marshaller mixin you inherit)
  rather than in the field. N sources means a mixin per combination — which is why there
  were five hand-rolled ones.
- Collisions have to be resolved by a precedence rule (`overrides` wins), and that rule
  has to be remembered, documented and tested. A body key that shadows a path parameter
  is a real risk — see the deleted `ViewArgAndJsonBodyMarshaller` — rather than something
  the type system makes unrepresentable.
- Models mix `BaseModel` with a marshaller mixin, so expressing `cls` correctly needs a
  `Protocol` and two `TypeVar`s.
- Errors come back in three shapes: pydantic's error list, `{"error": ...}`, and
  werkzeug's plain-text sentence served under an `application/json` mimetype.

The fix is to stop merging: let each source be named, and let path parameters arrive as
ordinary view-function arguments. FastAPI does this per field (`Path(...)`, `Query(...)`,
`Body(...)`); `flask-pydantic` does it by reading the view's type annotations.

### Why we are not doing that here

We are on **Flask 1.1.4** and **pydantic 1.10.26**, which excludes every current option:

| Library | Requires | Usable today |
|---|---|---|
| `webargs` 8.7 | Flask ≥ 0.12.5, marshmallow 3 | Yes — but marshmallow, a second validation library beside pydantic |
| `flask-pydantic` 0.11 | pydantic ≥ 1.7 | Yes — but it is the last pydantic-v1 release, Sept 2022 |
| `flask-pydantic` 0.12+ | pydantic ≥ 2 | No |
| `spectree` 2.0 | Flask ≥ 2, pydantic ≥ 2.11 | No |
| `apiflask` 3.1 | Flask ≥ 2.1, pydantic ≥ 2 | No |
| `flask-smorest` 0.47 | Flask ≥ 3.0.2, marshmallow | No |

So the hand-rolled marshallers are not a wrong turn so much as a reimplementation of a
library the stack cannot currently install.

### Options, roughly in order of lift

**1. Flask 1.1.4 → 3.x.** Routes and views barely move — `@blueprint.route` is stable
across the majors. The cost is dependency bumps, a handful of removed APIs
(`before_first_request`, `JSONEncoder` → the `app.json` provider), Werkzeug's majors, and
test-fixture plumbing. Note that Flask 2.1 changed `get_json()` to raise 415 rather than
return `None` for a non-JSON content type — `json_object_body` is insulated because it
checks `is_json` first, but the bare `get_json()` calls in `views/queries.py` and
`views/saved_queries.py` are not. Unblocks every library in the table above. Flask 1.1.4
has also been out of security support for years, which is an argument for this on its own.

**2. pydantic 1 → 2.** Independent of the above and independently shippable, but touches
every schema in the repo. Required for `flask-pydantic` 0.12+, `spectree` and `apiflask`.

**3. Rewrite in FastAPI.** Gets the per-field source declaration natively, and it is the
right destination *if* async I/O in `ui_api` is the goal. But it bundles: all 50 routes
rewritten; `lib/auth.py` rewritten, since identity is stashed as attributes on Flask's
request object (`request.claims`, `request.current_user`) and would become dependencies,
changing every view's decorator stack; ~348 test call sites moved off the Flask test
client; `OspreyFlask.make_response` reimplemented, including the `bigint_as_string`
encoding that exists because JS loses integer precision above 2^53; gunicorn → uvicorn;
and pydantic 2 as a hard requirement.

It also opens a question the codebase has so far avoided. `ui_api` is served by gunicorn's
gevent worker (`entrypoint.sh`), and is almost entirely storage-bound — nearly every
request reads or writes `osprey.worker.lib.storage.*`, which is synchronous SQLAlchemy
with a `scoped_session`. Notably, `osprey_async_worker` — the asyncio rewrite of the
event-processing path — contains no SQLAlchemy or storage code at all. The async work has
routed around exactly the layer `ui_api` is built on, so an async `ui_api` would have to
open that separately, and it could not be solved inside `ui_api` alone.

**Suggested order: 1, then adopt a validation library, then 2 if wanted.** Three
independently shippable, independently revertible steps that arrive at per-field source
declaration without a rewrite. Option 3 bundles all of it plus a concurrency-model change
into one commit that cannot be partially reverted.

Either way, `json_object_body` is throwaway in the good sense: whichever path is taken, it
is deleted in favour of the library's equivalent rather than needing to be unpicked.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of JSON request bodies across the UI API.
  - Empty or bodyless requests now use safe defaults where applicable.
  - GET, HEAD, and DELETE request bodies are ignored.
  - Invalid, missing, malformed, or non-object JSON bodies return clear 400 errors instead of unexpected server errors.
  - URL-provided values now correctly take precedence over conflicting body values.
  - Error responses no longer expose request content.

### DIFF
--- a/osprey_worker/src/osprey/worker/ui_api/osprey/lib/marshal.py
+++ b/osprey_worker/src/osprey/worker/ui_api/osprey/lib/marshal.py
@@ -12,12 +12,98 @@ from typing_extensions import Protocol
 T = TypeVar('T', bound='Marshallable')
 
 
+class UnreadableRequestBody(Exception):
+    """The request didn't arrive as JSON this API can read.
+
+    Raised by `json_object_body`, mapped to a 400 by `marshal_with`, so a marshaller
+    stays out of the business of building responses.
+
+    `message` is what the client is told, and it is a class attribute rather than a
+    constructor argument so that every string a caller can see is written here. Sending
+    `str(err)` instead would put the text under the control of whichever raise site ran,
+    and the natural thing to interpolate at a raise site is something off the request --
+    which is how an error message becomes a way to read server state back out. The same
+    reasoning keeps the configured path off `RulesDirNotADirectory`'s message in
+    `lib/rule_deployment`, where CodeQL flagged it as code-scanning/26.
+
+    Subclassed per failure so the client learns which of the two it hit without any of
+    them being assembled from the request.
+    """
+
+    message = 'Request body could not be read as JSON.'
+
+
+class RequestBodyNotDeclaredJson(UnreadableRequestBody):
+    message = 'A request body must be sent with Content-Type: application/json.'
+
+
+class RequestBodyNotAnObject(UnreadableRequestBody):
+    message = 'Request body must be a JSON object.'
+
+
+# The methods that carry no body, so nothing sent on one is read. Written as the
+# bodyless set rather than the submitting one so the default is to check: a method not
+# named here is treated as able to carry a body, and anything it sends has to hold up.
+_BODYLESS_METHODS = frozenset({'GET', 'HEAD', 'DELETE'})
+
+
+def json_object_body(flask_request: Request) -> dict[str, Any]:
+    """The request's JSON object body, or `{}` if it carried none.
+
+    Every marshaller that reads a body reads it through here, because "what counts as a
+    body" is the same question every time and it is easy to answer differently by
+    accident. It used to be answered separately in four places, which between them
+    dropped unlabelled bodies on the floor, returned 200 for bodies that never parsed,
+    and raised `TypeError` -- a 500 -- on a body that was a JSON list.
+
+    The difficulty is that `get_json` answers `None` to three different questions: a
+    body that wasn't labelled `application/json`, a body that didn't parse, and no body
+    at all. Reading `None` as "the caller sent nothing" collapses the first two into the
+    third, so a request whose body was dropped is answered as though it had asked for
+    the defaults -- and the endpoint reports success for work it did not do.
+
+    So the question asked here is whether bytes arrived, not which method carried them.
+    If they did, they must be declared JSON and must parse to an object. If they didn't,
+    the model's own fields decide whether that was allowed: a field with a default is
+    satisfied, and a required one produces pydantic's own error without this function
+    having an opinion about it.
+
+    GET, HEAD and DELETE carry no body, so anything sent on one is not read at all.
+    """
+    if flask_request.method in _BODYLESS_METHODS or not flask_request.get_data().strip():
+        return {}
+
+    if not flask_request.is_json:
+        raise RequestBodyNotDeclaredJson()
+
+    body = flask_request.get_json()
+    if not isinstance(body, dict):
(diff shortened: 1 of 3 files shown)


ThisIsMissEm:
@coderabbitai does your nitpick above still apply? It should be fixed now.

PR #508, opened 2026-09-17
Allow zooming out further in the Rules Visualizer

austin-ayers:
## Summary

The Rules Visualizer's zoom-out limit was pinned to whatever level the initial whole-graph auto-fit happened to land on, so there was no way to zoom out any further. `fit` only guarantees the whole graph's bounding box fits in the viewport — it says nothing about whether individual node labels overlap each other, so a dense graph could render with visibly colliding labels and no way to shrink further to fix it (zooming out shrinks nodes/labels together, which is the actual fix for overlap, at the cost of being harder to read at the extreme end).

- Relax the zoom-out floor to 30% of the initial fit level (`zoomUtils.ts`), giving room to shrink dense graphs further.
- Replace the standalone recenter button with a combined zoom control: `[ − ]  [ zoom% ]  [ + ]  |  [ recenter ]`, each zoom step ±20%, zoom centered on the viewport.
- Zoom percentage is shown relative to the initial fit ("100%" = whole graph fits), not Cytoscape's raw zoom units — previously a small graph could report a confusing number like 889%, since Cytoscape's zoom=1 means literal 1:1 pixel scale, not "fit to screen."

### Scope note

Implementing the above surfaced two bugs in this same code path, fixed here rather than split out, since they'd otherwise make the new zoom feature behave incorrectly:

1. **Pre-existing: recenter didn't return to the original 100% view (off by ~4%).** The initial layout fit and the recenter button's fit already used different padding (30px vs 0px) before this PR — just latent, since nothing previously gave a precise "100%" readout to notice the mismatch against. Both now share a single `GRAPH_FIT_PADDING` constant.
2. **Introduced by this PR: the zoom-percentage display fought the user ("rubberbanding" — snapping back to 100% on every scroll tick).** This one isn't pre-existing — it's a bug in the new code itself. Displaying a live zoom percentage means listening for Cytoscape's `zoom` event and calling `setState` on every tick, which re-renders `RulesVisualizerView`. That re-render exposed an existing (but previously never-triggered) fragility in `HierarchicalGraph`'s `memo()` wrapper, whose custom comparator compares `layoutOptions` (among other props) by reference equality. The new zoom-percentage code was passing a fresh `{ padding: ... }` object literal for `layoutOptions` on every render, which defeated that memo, so every zoom-triggered re-render destroyed and rebuilt the whole Cytoscape instance. Fixed by hoisting `layoutOptions` to a stable module-level reference. (Confirmed by hand this is a real infinite-render loop, not just a visual glitch — reverting the fix reproducibly OOM-crashes the process; see the regression test below.)

Also fixed while in here: the +/- zoom buttons were zooming around an arbitrary anchor point instead of the viewport center, which felt off. They now zoom centered.

## Testing

This PR also adds the first component-level tests in `osprey_ui` (previously the only test file covered a single pure function with no rendering) and wires `pnpm run test` into CI (`code-quality.yml`), since it wasn't run there before — so this repo's one existing test file wasn't actually enforced either.

- `zoomUtils.test.ts`: unit tests for the pure percentage/multiplier math.
- `RulesVisualizer.test.tsx`: mounts `RulesVisualizerView` with React Testing Library + jsdom, mocking the `cytoscape` module (it draws to `<canvas>`, which jsdom doesn't implement) so the actual wiring logic — not real rendering — is under test. Covers: loads at 100% and constructs the graph exactly once; **does not rebuild Cytoscape as the user zooms** (the regression test for bug #2 above — it includes a small circuit breaker that fails fast with a clear message if the graph gets rebuilt more than a few times, instead of letting a real regression hang/OOM the test run the way the original bug did); zoom percentage updates and recenter returns to exactly 100%.
- New devDependencies: `@testing-library/react`, `@testing-library/dom`, `jsdom` (the last was already present transitively via `@rstest/core`'s optional peer dependency; now declared explicitly since we depend on it directly).
- This is a new testing paradigm for `osprey_ui` (no prior use of `@testing-library` anywhere in the repo) — flagging in case reviewers want to weigh in on the approach before it becomes precedent for other components.

## Known limitations (not introduced by this PR)

- The visualizer doesn't re-fit or recompute on browser window resize — this is pre-existing behavior, not something introduced or fixed here.
- `HierarchicalGraph`'s memo comparator (deep-compares `elements` via `JSON.stringify`, reference-compares everything else, ignores `onLoad` entirely) is itself a bit fragile/inconsistent and was the root mechanism behind bug #2. Left as-is rather than refactored, to keep this PR's scope to the zoom feature — flagging as a possible follow-up.

## Test plan

- [ ] Open the Rules Visualizer, select a rule/label with a small graph — confirm it loads at 100% and the combined zoom/recenter control renders in the bottom-right corner.
- [ ] Select a large/dense graph — confirm you can scroll/pinch out noticeably further than before, and the percentage decreases below 100% as expected.
- [ ] Click `+`/`-` a few times — confirm the graph zooms centered on the viewport (not toward a corner), the percentage stays in sync, and it doesn't snap back to 100% mid-interaction.
- [ ] Click recenter after zooming/panning — confirm it snaps back to exactly 100%.
- [ ] `pnpm typecheck`, `pnpm test`, `pnpm lint` all pass locally (verified: 17/17 tests, 0 new lint warnings) and now run in CI.


austin-ayers:
Closing to keep iterating locally before this is ready for review.

